Issue #20, opened 2020-06-26
Add API to fetch all resources


PuneetPunamiya:
API would return all resources with their details and latest version sorted by rating and then by name.

Endpoint:
```
/resources?limit=<>
```
Limit - Returns the number of resources on the limit

PR #21, opened 2020-06-26
API to fetch all resources

PuneetPunamiya:
This API returns all resources with their details and latest version sorted by rating and then by name.

* Adds api structure for resource api
*  Resource api implementation and tests

Closes: #20


api/cmd/hub/main.go
```diff
@@ -101,15 +105,15 @@ func main() {
...
+			handleHTTPServer(ctx, u, categoryEndpoints, resourceEndpoints, &wg, errc, logger, *dbgF)
 		}
 
 	default:
 		fmt.Fprintf(os.Stderr, "invalid host argument: %q (valid hosts: localhost)\n", *hostF)
 	}
 
 	// Wait for signal.
-	logger.Infof("exiting (%v)", <-errc)
+	logger.Info("exiting (%v)", <-errc)
```
SM43, line 116: any reason this is changed?

api/design/resource.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// Method to get all resources
+	Method("AllResources", func() {
+		Description("Get all Resources")
+		Payload(func() {
+			Attribute("limit", UInt, "Number of resources", func() {
+				Default(100)
+			})
+		})
+
```
SM43, line 20: no need of blank line

api/design/resource.go
```diff
@@ -0,0 +1,30 @@
+package design
+
+import (
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("resource", func() {
+	Description("The resource service provides all resources information")
+
+	Error("db-down", ErrorResult)
```
SM43, line 10: db-down -> internal-error ?

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+
+	Attribute("id", UInt, "ID is the unique id of the resource", func() {
+		Example("id", 1)
+	})
+
+	Attribute("name", String, "Name of the resource", func() {
+		Example("name", "golang")
+	})
+
+	Attribute("displayName", String, "Display name of the resource", func() {
```
SM43, line 20: this should be included in version info

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+		Example("displayName", "golang")
+	})
+
+	Attribute("catalog", Catalog, "Type of catalog where resource belongs")
+
+	Attribute("type", String, "Type of resource", func() {
+		Example("type", "task")
+	})
+
+	Attribute("description", String, "Description of resource", func() {
```
SM43, line 30: this should also be included in version info

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+		Example("descripiton", "This Task is Golang task to build Go projects")
+	})
+
+	Attribute("latest_version", String, "Latest version o resource", func() {
```
SM43, line 34: `latest_version` can have the object of `version` which will include all info of that particular version - URLs , displayName , description

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+		Example("latest_version", "0.1")
+	})
+
+	Attribute("tags", ArrayOf(ResourceTag), "Tags related to resources")
+
+	Attribute("rating", UInt, "Rating of resource", func() {
```
SM43, line 40: rating will have `float` value

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+		Attribute("type")
+		Attribute("description")
+		Attribute("latest_version")
+		Attribute("tags")
+		Attribute("rating")
+		Attribute("last_updated_at")
+		Attribute("versions")
+	})
+
+	View("extended", func() {
```
SM43, line 64: this api will not need `view` as only one `resource` is needed

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+	Attribute("versions", ArrayOf(Versions), "Version of resource")
+
+	View("default", func() {
+		Attribute("id")
+		Attribute("name")
+		Attribute("displayName")
+		Attribute("catalog")
+		Attribute("type")
+		Attribute("description")
+		Attribute("latest_version")
```
SM43, line 57: can we use one under_scores or camelCase? React prefer camelCase I guess

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+		Attribute("tags")
+		Attribute("rating")
+		Attribute("last_updated_at")
+	})
+	Required("id", "name", "displayName", "catalog", "type", "description", "latest_version", "tags", "rating", "last_updated_at", "versions")
+
+})
+
+var Versions = Type("Versions", func() {
```
SM43, line 80: this can have description, displayName, urls

api/gen/http/openapi.yaml
```diff
@@ -79,6 +123,51 @@ definitions:
         description: Is the error a timeout?
         example: false
     description: All_internal-error_Response_Body result type (default view)
+    example:
+      fault: true
+      id: 123abc
```
SM43, line 128: 123abc? can you check why this has abc?

api/pkg/service/resource.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, r := range all {
+		tags := []*resource.Tag{}
+		for _, t := range r.Tags {
+			tags = append(tags, &resource.Tag{
+				ID:   t.ID,
+				Name: t.Name,
+			})
+		}
+		latestVersion := r.Versions[len(r.Versions)-1]
+		res = append(res, &resource.Resource{
```
SM43, line 51: can you write this as a separate function? for mapping DB models to return types. It can be reused.

api/pkg/service/resource.go
```diff
@@ -0,0 +1,66 @@
...
+
+	for _, r := range all {
+		tags := []*resource.Tag{}
+		for _, t := range r.Tags {
+			tags = append(tags, &resource.Tag{
+				ID:   t.ID,
+				Name: t.Name,
+			})
+		}
+		latestVersion := r.Versions[len(r.Versions)-1]
```
SM43, line 50: here we will have to sort by `version` rather than `id` ?? eg. `0.1.1` > `0.1`

api/fixtures/catalogs.yaml
```diff
@@ -0,0 +1,13 @@
+- id: 1
+  name: Tekton
+  type: official
+  url: https:/github.com/Pipelines-Marketplace/catalog
+  created_at: 2016-01-01 12:30:12
+  updated_at: 2016-01-01 12:30:12
+
+- id: 2
+  name: Tekton
```
SM43, line 9: same values for both?

api/pkg/service/resource_test.go
```diff
@@ -0,0 +1,22 @@
...
+	resourceSvc resource.Service
+)
+
+func Test_AllResources(t *testing.T) {
+	resourceSvc = NewResource(testConfig)
+	LoadFixture(db, "../../fixtures")
+	payload := &resource.AllResourcesPayload{Limit: 2}
+	all, err := resourceSvc.AllResources(context.Background(), payload)
+	assert.NoError(t, err)
+	assert.Equal(t, len(all), 2)
```
SM43, line 21: expected value should be in middle and actual on the right side.

api/design/type.go
```diff
@@ -4,6 +4,101 @@ import (
...
+	Attribute("versionId", UInt, "Version ID of the resource to be fetched", func() {
+		Example("versionId", 1)
+	})
+
+	Attribute("version", String, "Version of the resource to be fetched", func() {
+		Example("versionId", "0,1")
```
SM43, line 86: Example for version..

PR #30, opened 2020-07-08
Run postgres service instead of using DIND to run database

sthaha:
Run postgres service instead of using docker (DIND)

API tests require postgres to be running. Previously we were using
DIND (Docker In Docker) to run postgres as a container which was
quite flaky.

This patch addresses the flakiness by
    - installing  postgres
    - running the service in background
before the tests are run.

Signed-off-by: Sunil Thaha <sthaha@redhat.com>


vdemeester:
Given https://github.com/tektoncd/plumbing/pull/466 I think.. we can't do that :sob:

PR #32, opened 2020-07-09
Adds minPipelinesVersion field in ResourceVersion Table

SM43:
Changes:

This adds a database migration adding a column minPipelineVersionin ResourceVersion table which would tell the minimum pipelines version the resource is compatible with.

### DIFF
--- a/api/pkg/db/model/model.go
+++ b/api/pkg/db/model/model.go
@@ -43,12 +43,13 @@ type (
 
 	ResourceVersion struct {
 		gorm.Model
-		Version     string
-		Description string
-		URL         string
-		DisplayName string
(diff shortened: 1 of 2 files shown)


api/pkg/db/model/model.go
```diff
@@ -43,12 +43,13 @@ type (
...
-		Description string
-		URL         string
-		DisplayName string
-		Resource    Resource
-		ResourceID  uint
+		Version             string
+		Description         string
+		URL                 string
+		DisplayName         string
+		MinPipelinesVersion string
```
sthaha, line 50: shouldn't we make all the mandatory fields `not-null` ?

PR #33, opened 2020-07-09
Adds necessary constraints to database fields

PuneetPunamiya:
This adds necessary constraints to db fields and tests to validate them.
Also moves all migrations to their own files.


api/pkg/db/model/model.go
```diff
@@ -21,19 +21,19 @@ type (
 
 	Catalog struct {
 		gorm.Model
-		Name       string
-		Type       string
-		URL        string
-		Owner      string
+		Name       string `gorm:"size:100;not null;unique"`
```
sthaha, line 24: should we set a size?

api/pkg/db/model/model.go
```diff
@@ -21,19 +21,19 @@ type (
...
+		Type       string `gorm:"size:100;not null;unique"`
```
sthaha, line 25: why is the type unique?

api/pkg/db/model/model.go
```diff
@@ -21,19 +21,19 @@ type (
...
+		URL        string `gorm:"size:100;not null;unique"`
```
sthaha, line 26: why is this unique? wouldn't this prevent 2 branches of the same repo to be added a s catalog?

sthaha: IMHO, we should only add constraints that are absolutely required by the domain we are trying to map.

api/pkg/db/model/model.go
```diff
@@ -21,19 +21,19 @@ type (
...
+		Owner      string `gorm:"size:100;not null;unique"`
```
sthaha, line 27: I think catalog owner can also repeat, so `unique` isn't correct.

sthaha, line 27: Do we need this field? Can we remove this until a feature requires it? I am also thinking that the mapping may be one to many; i.e. a catalog is owned by multiple owners ..

api/pkg/db/model/model.go
SM43: i am confused whether to set or size, by default it is 255

api/pkg/db/model/model.go
```diff
@@ -21,19 +21,19 @@ type (
...
 		Resources  []Resource
-		Revision   string
+		Revision   string `gorm:"size:100;not null;unique"`
 	}
 
 	Resource struct {
 		gorm.Model
-		Name      string
-		Type      string
+		Name      string `gorm:"size:100;not null;unique"`
```
sthaha, line 35: name isn't unique but `catalog - name` combination is

api/pkg/db/model/model.go
```diff
@@ -21,19 +21,19 @@ type (
...
+		Type      string `gorm:"size:100;not null;unique"`
```
sthaha, line 36: it isn't unique; see above

api/pkg/db/model/model.go
SM43, line 35: resource name should not be unique, can repeat. we can have multiple catalogs

api/pkg/db/model/model.go
SM43, line 36: Type also can't be unique

api/pkg/db/model/model.go
```diff
@@ -43,11 +43,11 @@ type (
 
 	ResourceVersion struct {
 		gorm.Model
-		Version             string
+		Version             string `gorm:"size:100;not null;unique"`
```
SM43, line 46: Version can't be unique eg. all the resources will have `v0.1`

api/pkg/db/model/model.go
```diff
@@ -43,11 +43,11 @@ type (
...
 		Description         string
-		URL                 string
+		URL                 string `gorm:"size:100;not null;unique"`
 		DisplayName         string
-		MinPipelinesVersion string
+		MinPipelinesVersion string `gorm:"size:100;not null;unique"`
```
SM43, line 50: MinPipelinesVersion will repeat so can't be unique

api/cmd/db/migration.go
```diff
@@ -18,6 +18,54 @@ func Migrate(api *app.APIConfig) error {
 		api.DB(),
 		gormigrate.DefaultOptions,
 		[]*gormigrate.Migration{
+			{
+				ID: "202007141723",
+				Migrate: func(tx *gorm.DB) error {
+
```
SM43, line 24: All queries can be written under one migration

api/cmd/db/migration.go
```diff
@@ -122,14 +170,14 @@ func fkey(log *zap.SugaredLogger, db *gorm.DB, model interface{}, args ...string
...
 	var categories = map[string][]string{
-		"Others":         []string{},
-		"Build Tools":    []string{"build-tool"},
-		"CLI":            []string{"cli"},
-		"Cloud":          []string{"gcp", "aws", "azure", "cloud"},
-		"Deploy":         []string{"deploy"},
-		"Image Build":    []string{"image-build"},
-		"Notification":   []string{"notification"},
-		"Test Framework": []string{"test"},
+		"Others":         {},
```
SM43, line 173: why this change?

api/pkg/db/model/model.go
```diff
@@ -7,33 +7,33 @@ import (
...
 	Catalog struct {
 		gorm.Model
-		Name       string
-		Type       string
-		URL        string
-		Owner      string
+		Name       string `gorm:"not null"`
+		Type       string `gorm:"not null"`
+		URL        string `gorm:"not null"`
+		Owner      string `gorm:"not null"`
```
SM43, line 27: you can remove this field as discussed before.

SM43: Owner

api/pkg/db/model/model.go
```diff
@@ -7,33 +7,33 @@ import (
...
 		Resources  []*Resource `gorm:"many2many:resource_tags;"`
 	}
 
 	Catalog struct {
 		gorm.Model
-		Name       string
-		Type       string
-		URL        string
-		Owner      string
+		Name       string `gorm:"not null"`
```
SM43, line 24: `catalog - name - type` combination will be unique if you can add a constraint for that, would be nice

api/pkg/db/migration/migration_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func TestMigrateCheckNotNull(t *testing.T) {
+	tc := testutils.Setup(t)
+
+	err := tc.Error()
+
+	db := tc.DB()
+
+	err = db.Create(&model.Catalog{Name: "tekton", Type: "", URL: "", Revision: "master"}).Error
```
SM43, line 18: `assert` should be for each query

api/pkg/db/migration/migration_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/tektoncd/hub/api/pkg/db/model"
+	"github.com/tektoncd/hub/api/pkg/testutils"
+)
+
+func TestMigrateCheckNotNull(t *testing.T) {
+	tc := testutils.Setup(t)
+
+	err := tc.Error()
```
SM43, line 14: why this is needed?

api/pkg/db/model/model.go
```diff
@@ -21,33 +21,32 @@ import (
 type (
 	Category struct {
 		gorm.Model
-		Name string `gorm:"size:100;not null;unique"`
+		Name string `gorm:"not null"`
```
SM43, line 24: There can't be 2 categories with same name. So, unique can be applied here.

api/pkg/db/model/model.go
```diff
@@ -21,33 +21,32 @@ import (
...
 		Tags []Tag
 	}
 
 	Tag struct {
 		gorm.Model
-		Name       string `gorm:"size:100;not null;unique"`
+		Name       string `gorm:"not null"`
```
SM43, line 30: Similarly here also there can't be 2 tags with same name. So, unique can be applied here.

SM43:
@PuneetPunamiya There is no Migration written. Does this case not need one?

PuneetPunamiya:
> @PuneetPunamiya There is no Migration written. Does this case not need one?

Done !

api/pkg/db/model/model.go
```diff
@@ -21,33 +21,32 @@ import (
...
 		CategoryID uint
 		Resources  []*Resource `gorm:"many2many:resource_tags;"`
 	}
 
 	Catalog struct {
 		gorm.Model
-		Name       string
-		Type       string
-		URL        string
-		Owner      string
```
SM43, line 41: Removing this field will not delete the column, you will need to delete in the migration

api/cmd/db/migration.go
```diff
@@ -32,6 +32,50 @@ func Migrate(api *app.APIConfig) error {
...
+					}
+					if err := tx.Exec(resource_version_query).Error; err != nil {
+						log.Error(err)
+						return err
+					}
+					return nil
+				},
+			},
+			{
+				ID: "202005080741",
```
SM43, line 70: Why a separate migration for this?

api/cmd/db/migration.go
```diff
@@ -32,6 +32,44 @@ func Migrate(api *app.APIConfig) error {
 		api.DB(),
 		gormigrate.DefaultOptions,
 		[]*gormigrate.Migration{
+			{
+				ID: "202007141723",
```
SM43, line 36: Could you please add a comment above `ID` what the migration does?

api/pkg/db/model/model.go
```diff
@@ -21,33 +21,32 @@ import (
...
+		Name       string `gorm:"primary_key"`
+		Type       string `gorm:"primary_key"`
+		URL        string `gorm:"not null"`
```
SM43, line 40: Does this not requires `default:null`?

api/pkg/db/validation/validation_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/tektoncd/hub/api/pkg/db/model"
+	"github.com/tektoncd/hub/api/pkg/testutils"
+)
+
+func TestCheckNotNull(t *testing.T) {
+	tc := testutils.Setup(t)
+
+	err := tc.Error()
```
SM43, line 30: Is this required here?

api/pkg/db/validation/validation_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	db := tc.DB()
+
+	err = db.Create(&model.Catalog{Name: "tekton", Type: "", URL: "", Revision: "master"}).Error
```
SM43, line 34: I think this will return an error only because `Type` is null and we won't know whether `URL` is `NOT NULL` or not.

api/pkg/db/model/model.go
```diff
@@ -21,33 +21,32 @@ import (
...
 		ContextDir string
 		Resources  []Resource
-		Revision   string
+		Revision   string `gorm:"not null"`
```
SM43, line 43: Using only `NOT NULL` won't be helpful I guess, we also need `default:null`? WDYT?

api/pkg/db/initializer/initiailizer.go
```diff
@@ -81,7 +81,6 @@ func (i *Initializer) addCatalogs() error {
 			Name:       c.Name,
 			Type:       c.Type,
 			URL:        c.URL,
-			Owner:      c.Owner,
```
SM43, line 84: Could you also remove `owner` from config.yaml and `initializer`?

api/pkg/db/migration/202006071000_add_foreign_Key_constraints.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/jinzhu/gorm"
+	"go.uber.org/zap"
+	"gopkg.in/gormigrate.v1"
+
+	"github.com/tektoncd/hub/api/pkg/db/model"
+)
+
+func addForeignKeyConstraints(log *zap.SugaredLogger) *gormigrate.Migration {
```
SM43, line 25: this migration adds the resource table and then adds the `fk` so I think the name should be something related to creating tables.

api/pkg/db/migration/202008071700_add_migration_on_catalog.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"github.com/jinzhu/gorm"
+	"go.uber.org/zap"
+	"gopkg.in/gormigrate.v1"
+
+	"github.com/tektoncd/hub/api/pkg/db/model"
+)
+
+func addMigrationOnCatalog(log *zap.SugaredLogger) *gormigrate.Migration {
```
SM43, line 25: I think Migration don't need to be in the name, could be what the migration does?

PR #47, opened 2020-07-23
Adds Config file to populate application data in DB

SM43:
Previously, while running migration to create tables we used to insert data
into tags, categories and catalogs tables but now we keep that data in config
file at root of repo and reads from that file and add to db while starting
the api server.


api/.env.dev
```diff
@@ -5,3 +5,5 @@ POSTGRESQL_PORT="5432"
 POSTGRESQL_DATABASE="hub"
 POSTGRESQL_USER="postgres"
 POSTGRESQL_PASSWORD="postgres"
+
+CONFIG_FILE_URL="https://raw.githubusercontent.com/sm43/hub/read-config-from-repo/config.yaml"
```
PuneetPunamiya, line 9: Should this yaml point to your repo ??

api/design/status.go
```diff
@@ -1,3 +1,17 @@
+// Copyright © 2020 The Tekton Authors.
```
PuneetPunamiya, line 1: Should this change come under this commit ?

sthaha: yes, for all new files

api/.env.dev
sthaha, line 9: I think it is better to point to the `file://../config.yaml` ?

api/cmd/api/main.go
```diff
@@ -63,6 +67,27 @@ func main() {
...
+		// Read the config file from the repository to add into the DB
+		data, err := app.ReadConfig(api.ConfigFileURL())
+		if err != nil {
+			logger.Fatalf("Failed to read config file: %v", err)
+		}
+
+		// Viper instance to read data from file
+		v := viper.New()
+		v.SetConfigType("yaml")
+		v.ReadConfig(bytes.NewBuffer(data))
```
sthaha, line 80: I feel this code should move into api.Config itself

api/cmd/api/main.go
```diff
@@ -63,6 +64,14 @@ func main() {
 		defer api.Cleanup()
 	}
 
+	{
```
sthaha, line 67: we don't need the `{ } ` do we?

api/cmd/api/main.go
```diff
@@ -63,6 +64,14 @@ func main() {
...
+		// Populate Tables
+		initializer := initializer.New(api)
```
sthaha, line 69: should we skip this in `dev` mode :) ?

SM43: we will address this later as discussed.

api/config/10-db-secret.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
     app: db
 type: Opaque
 stringData:
-  POSTGRESQL_DATABASE: hub
-  POSTGRESQL_USER: postgres
-  POSTGRESQL_PASSWORD: postgres
-  POSTGRESQL_PORT: "5432"
+  POSTGRES_DB: hub
```
sthaha, line 24: why is this change needed?

api/config/21-api-deployment.yaml
```diff
@@ -53,25 +53,30 @@ spec:
             successThreshold: 1
             timeoutSeconds: 1
           env:
-            - name: POSTGRESQL_HOST
+            - name: POSTGRES_HOST
```
sthaha, line 56: isn't this an unrelated change?

api/pkg/app/app.go
```diff
@@ -218,3 +240,14 @@ func initLogger(mode EnvMode) (*zap.SugaredLogger, error) {
 	}
 	return log.Sugar(), nil
 }
+
+// configFileURL will look for CONFIG_FILE_URL to be defined among
+// environment variables
+func configFileURL() (string, error) {
+
+	if val := viper.GetString("CONFIG_FILE_URL"); val != "" {
+		return val, nil
```
sthaha, line 249: could change this to 
```
val = viper.GetString ..
if  val == "" {
   return "", error 
}
return val, nil
```
so that the exception is inside the if

api/pkg/db/initializer/initiailizer.go
```diff
@@ -0,0 +1,92 @@
...
+
+	for _, c := range i.data.Categories {
+		cat := &model.Category{Name: c.Name}
+		if err := i.db.Where(&cat).FirstOrCreate(&cat).Error; err != nil {
+			i.log.Error(err)
+			return err
+		}
+		for _, t := range c.Tags {
+			tag := &model.Tag{Name: t, CategoryID: cat.ID}
+			if err := db.Model(&model.Tag{}).Where(&model.Tag{Name: t}).FirstOrCreate(&tag).Error; err != nil {
```
sthaha, line 59: how about breaking this long line into 

```
err := db.Model....;
if err != nil {
}
```

api/pkg/db/initializer/initiailizer.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, c := range i.data.Catalogs {
+		cat := &model.Catalog{
+			Name:       c.Name,
+			Type:       c.Type,
+			URL:        c.URL,
+			Owner:      c.Owner,
+			Revision:   c.Revision,
+			ContextDir: c.ContextDir,
+		}
+		if err := db.Where(&model.Catalog{Name: c.Name, Type: c.Type}).FirstOrCreate(&cat).Error; err != nil {
```
sthaha, line 86: same, could we break the long line?

api/pkg/db/model/model.go
```diff
@@ -29,7 +29,7 @@ type (
 		gorm.Model
 		Name       string `gorm:"size:100;not null;unique"`
 		Category   Category
-		CategoryID int
+		CategoryID uint
```
sthaha, line 32: does this need any migration changes? I guess not but could you please confirm?

api/test/config/env.test
```diff
@@ -1,6 +1,9 @@
 ENVIRONMENT="test"
-POSTGRESQL_HOST="localhost"
-POSTGRESQL_PORT="5432"
-POSTGRESQL_DATABASE="hub_test"
-POSTGRESQL_USER="postgres"
-POSTGRESQL_PASSWORD="postgres"
+
```
sthaha, line 2: unrelated change ?

api/test/config/env.test
```diff
@@ -1,6 +1,9 @@
...
+POSTGRES_HOST="localhost"
+POSTGRES_PORT="5432"
+POSTGRES_DB="hub_test"
+POSTGRES_USER="postgres"
+POSTGRES_PASSWORD="postgres"
+
+CONFIG_FILE_URL="file://../../../../config.yaml"
```
sthaha, line 9: its better to keep another config.yaml file for test, that can be empty

test/presubmit-tests.sh
```diff
@@ -57,14 +57,14 @@ set-pg-passwd() {
 api-unittest(){
   install-postgres
   source $API_DIR/test/config/env.test
-  set-pg-passwd "$POSTGRESQL_PASSWORD"
+  set-pg-passwd "$POSTGRES_PASSWORD"
```
sthaha, line 60: unrelated change.

api/pkg/db/model/model.go
SM43: Ideally yes it needs a migration but it won't be breaking anything because `CategoryID` refers to Tag `Id` which is the primary key and autogenerated and won't be a signed int. And we don't access category Id except for preloading.

SM43: but yes will write a migration anyway. Thanks for reminding :)

SM43: I checked this one, queries are the same for `uint` and `int` as Postgres have the `integer` data type. so, skipping the migration.

api/test/config/env.test
```diff
@@ -1,6 +1,9 @@
...
-POSTGRESQL_USER="postgres"
-POSTGRESQL_PASSWORD="postgres"
+
+POSTGRES_HOST="localhost"
+POSTGRES_PORT="5432"
+POSTGRES_DB="hub_test"
+POSTGRES_USER="postgres"
+POSTGRES_PASSWORD="postgres"
+
+CONFIG_FILE_URL="file://../../../test/config/config.yaml"
```
sthaha, line 9: @SM43 could you please use a different file config.yaml file for test?

api/config/20-api-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: api
+  namespace: tekton-hub
+  labels:
+    app: api
+data:
+  CONFIG_FILE_URL: https://raw.githubusercontent.com/tektoncd/hub/master/config.yaml
```
sthaha, line 23: is this a valid yaml?

SM43: Once the pr is mergered then it will be

api/pkg/db/initializer/initiailizer.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+func (i *Initializer) addCategories() error {
+
+	db := i.db
+
+	// Checks if tables exists
+	if !db.HasTable(&model.Category{}) || !db.HasTable(model.Tag{}) {
+		return fmt.Errorf("Categories or Tags Table not found")
```
sthaha, line 48: could you please use lowercase of `errors` 🙏

PR #54, opened 2020-07-29
Adds readiness and liveness probe in db deployment

SM43:
# Changes
Adds readiness and liveness probe in db deployment

# Submitter Checklist

review them:

### DIFF
--- a/api/config/12-db-deployment.yaml
+++ b/api/config/12-db-deployment.yaml
@@ -57,6 +57,16 @@ spec:
           volumeMounts:
             - name: db
               mountPath: /var/lib/postgresql/data
+          readinessProbe:
+            exec:
+              command: [bash, -c ,"psql -w -U ${POSTGRES_USER} -d ${POSTGRES_DB} -c 'SELECT 1'"]
+            initialDelaySeconds: 15
+            timeoutSeconds: 2
+          livenessProbe:
+            exec:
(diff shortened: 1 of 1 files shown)


api/config/12-db-deployment.yaml
```diff
@@ -57,6 +57,16 @@ spec:
           volumeMounts:
             - name: db
               mountPath: /var/lib/postgresql/data
+          readinessProbe:
+            exec:
+              command: [psql, -w, -U, postgres, -d, $(POSTGRES_DB), -c, "SELECT 1"]
```
sthaha, line 62: does this work? what does $(POSTGRES_DB) eval to?

sthaha, line 62: if the variable interpolation doesn't work, how about we use `bash -c psql .... ` ?

SM43: `$(POSTGRES_DB)` refers value from env variable defined above

sthaha:
```  
┆ ┆ ┆ ┆ ┆ - bash
  ┆ ┆ ┆ ┆ ┆ - -c
  ┆ ┆ ┆ ┆ ┆ - psql -w -U ${POSTGRES_USER} -d ${POSTGRES_DB} -c 'select 1 as query;'
  ```
works with env

SM43: It works. Thank you :)

PR #58, opened 2020-08-10
 Add Org Column & drops Owner column from Catalog Table

SM43:
# Changes
Adds org column which would be org name of the catalog repo.
Owners will be added later as a array of user who will be authorized
to sync the catalog with Hub.

# Submitter Checklist

review them:

### DIFF
--- a/api/cmd/db/migration.go
+++ b/api/cmd/db/migration.go
@@ -33,6 +33,7 @@ func Migrate(api *app.APIConfig) error {
 		gormigrate.DefaultOptions,
 		[]*gormigrate.Migration{
 			{
+				// Creates Resource, ResourceVersion & Catalog tables and foreign keys on them
 				ID: "202006071000",
 				Migrate: func(tx *gorm.DB) error {
 
@@ -58,6 +59,7 @@ func Migrate(api *app.APIConfig) error {
 				},
 			},
 			{
+				// Adds minPipelineVersion Column in ResourceVersion Table
 				ID: "202006091100",
 				Migrate: func(tx *gorm.DB) error {
 					if err := tx.AutoMigrate(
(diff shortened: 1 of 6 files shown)


api/test/fixtures/catalogs.yaml
```diff
@@ -13,20 +13,20 @@
 # limitations under the License.
 
 - id: 1
-  name: Tekton
+  name: catalog-official
```
PuneetPunamiya, line 16: Should it be just `name: official` ?

api/cmd/db/migration.go
```diff
@@ -33,6 +33,49 @@ func Migrate(api *app.APIConfig) error {
...
+						log.Error(err)
+						return err
+					}
+					if err := tx.Model(&model.Catalog{}).
+						AddUniqueIndex("idx_name_org_code", "name", "org").Error; err != nil {
+						log.Error(err)
+						return err
+					}
+
+					catalogQuery := `ALTER TABLE catalogs
```
sthaha, line 57: wouldn't automigrate create these?

api/pkg/db/model/model.go
```diff
@@ -35,13 +35,13 @@ type (
...
-		Name       string
-		Type       string
-		URL        string
-		Owner      string
-		ContextDir string
+		Name       string `gorm:"unique_index:idx_name_org_code"`
+		Org        string `gorm:"unique_index:idx_name_org_code"`
+		Type       string `gorm:"not null;default:null"`
+		URL        string `gorm:"not null;default:null"`
+		ContextDir string `gorm:"not null;default:null"`
```
sthaha, line 42: I am not sure if we need "not null". Also why is the default: null when there is a constraint called `not null` 🤔

config.yaml
```diff
@@ -31,9 +31,9 @@ categories:
...
 
 catalogs:
-  - name: tektoncd
+  - name: catalog
+    org: tektoncd
     type: community
-    owner: tektoncd
     url: https://github.com/tektoncd/catalog
     revision: master
     contextDir: /
```
sthaha, line 39: this is supposed to be optional and treated as "" by default.

api/cmd/db/migration.go
SM43: automigrate does not update constraints
 http://gorm.io/docs/migration.html

config.yaml
SM43: Yeah. I think we should remove this field from here and at the bottom of config file mention all fields their explanation

api/pkg/db/model/model.go
SM43: `not null` is applied on table but when we create a record using `db.Create` if a field is not mentioned in struct, gorm adds 
`""` by default.
 So, for example if `url` is not defined then ideally it should return error but as gorm add `""`, it does not return error. due to which we add default as null and expects some value.

config.yaml
sthaha: +1

Issue #57, opened 2020-08-04
Add a API for user login using a GitHub Oauth


SM43:
Adds an Endpoint `/auth/login` which will authenticate a user using GitHub OAuth,
add the user in DB, and returns a jwt which would be needed to access rating APIs.

PR #59, opened 2020-08-10
Add a API for user login using a GitHub Oauth

SM43:
# Changes
Adds an Endpoint `/auth/login` which will authenticate a user using GitHub OAuth, add the user in DB,
and returns a jwt which would be needed to access rating APIs.

Issue #57

# Submitter Checklist

review them:


api/design/auth.go
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+package design
+
+import (
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("auth", func() {
+	Description("The auth service exposes endpoint to authenticates User using GitHub Oauth")
```
PuneetPunamiya, line 22: Can we change the description to `Authenticates users using Github Auth`

api/.env.dev
```diff
@@ -6,4 +6,8 @@ POSTGRES_DB="hub"
 POSTGRES_USER="postgres"
 POSTGRES_PASSWORD="postgres"
 
-CONFIG_FILE_URL="file://../config.yaml"
\ No newline at end of file
+CONFIG_FILE_URL="file://../config.yaml"
+
+GH_CLIENT_ID=""
```
PuneetPunamiya, line 11: Can we move this above `CONFIG_FILE_URL`?

api/design/auth.go
```diff
@@ -0,0 +1,48 @@
...
+		HTTP(func() {
+			POST("/auth/login")
+			Param("code")
+
+			Response(StatusOK)
+			Response("invalid-code", StatusBadRequest)
+			Response("internal-error", StatusInternalServerError)
+		})
+	})
+
```
PuneetPunamiya, line 47: Can you remove this blank line please ?

api/test/config/env.test
```diff
@@ -6,4 +6,8 @@ POSTGRES_DB="hub_test"
 POSTGRES_USER="postgres"
 POSTGRES_PASSWORD="postgres"
 
-CONFIG_FILE_URL="file://../../../test/config/config.yaml"
\ No newline at end of file
+CONFIG_FILE_URL="file://../../../test/config/config.yaml"
+
+GH_CLIENT_ID="test-client-id"
```
PuneetPunamiya, line 11: Can we move this above ??

api/pkg/service/auth/auth_http_test.go
```diff
@@ -0,0 +1,113 @@
...
+		server.MountLoginHandler,
+		auth.NewLoginEndpoint(New(tc)))
+	return checker
+}
+
+func TestLogin_Http(t *testing.T) {
+	tc := testutils.Setup(t)
+	testutils.LoadFixtures(t, tc.FixturePath())
+
+	defer gock.Disable()
```
PuneetPunamiya, line 46: Can we create a function for gock as it used in both the tests i.e. valid and invalid code ?

SM43: but the code is different for both the test..

SM43: response for mock is different

PuneetPunamiya: Okay got it :+1:

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	// adds user in db if not exist
+	user, err := s.addUser(ghUser)
+	if err != nil {
+		return nil, err
+	}
+
+	// gets user scopes to add in jwt
+	scopes, err := s.getUserScopes()
```
sthaha, line 74: should be `defaultScopes` - right?

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,141 @@
...
+	var scopes []model.Scope
+	s.db.LogMode(true)
+	if err := s.db.Where(&model.Scope{Default: true}).
+		Find(&scopes).
+		Error; err != nil {
+		s.logger.Error(err)
+		return nil, internalError
+	}
+
+	var userScopes []string
```
sthaha, line 117: could you please use `gorm -> Pluck` for this?

api/pkg/service/auth/auth_http_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	gock.New("https://api.github.com").
+		Get("/user").
+		Reply(200).
+		JSON(map[string]string{
+			"login": "test",
+			"name":  "test-user",
+		})
+
+	var token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9." +
```
sthaha, line 67: could you please add a comment about the magic string :)

api/pkg/service/auth/auth_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	gock.New("https://api.github.com").
+		Get("/user").
+		Reply(200).
+		JSON(map[string]string{
+			"login": "test",
+			"name":  "test-user",
+		})
+
+	var token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9." +
```
sthaha, line 50: do you think we can use a package level `const ValidToken string = ".... "` for this?

api/design/auth.go
```diff
@@ -0,0 +1,47 @@
...
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("auth", func() {
+	Description("The auth service exposes endpoint to authenticates User using GitHub Oauth")
+
+	Error("invalid-code", ErrorResult, "Invalid Authorization code")
+	Error("internal-error", ErrorResult, "Internal Server Error")
+
+	Method("Login", func() {
```
sthaha, line 27: could we call this method -  authenticate?

api/pkg/app/app.go
```diff
@@ -252,3 +312,34 @@ func configFileURL() (string, error) {
 	}
 	return val, nil
 }
+
+// initGitHub looks for configuration among environment variables
+// and intialises the GitHub Oauth Config
+func initOauthConfig() (*oauth2.Config, error) {
```
sthaha, line 318: nit: OAuth instead of `Oauth`

api/pkg/db/initializer/initiailizer.go
```diff
@@ -19,7 +19,7 @@ type Initializer struct {
 }
 
 // New returns the Initializer implementation.
-func New(c app.Config) *Initializer {
+func New(c app.Base) *Initializer {
```
sthaha, line 22: `app.Base` reads odd. How about `app.MinConfig` ? or `app.BaseConfig` ?

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/tektoncd/hub/api/gen/auth"
+	"github.com/tektoncd/hub/api/pkg/app"
+	"github.com/tektoncd/hub/api/pkg/db/model"
+)
+
+type service struct {
+	logger        *zap.SugaredLogger
+	db            *gorm.DB
+	conf          *oauth2.Config
```
sthaha, line 35: could you call this var `oauth` instead of `conf` ?

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+func (s *service) Login(ctx context.Context, p *auth.LoginPayload) (res *auth.LoginResult, err error) {
+
+	// gets access_token for user using authorization_code
+	token, err := s.conf.Exchange(oauth2.NoContext, p.Code)
+	if err != nil {
+		return nil, invalidCode
+	}
+
+	// gets user details from github using the access_token
+	oauthClient := s.conf.Client(oauth2.NoContext, token)
```
sthaha, line 60: It reads better if it said 
```
oauthClient = s.oauth.Client(...)
```

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+	ghClient := github.NewClient(oauthClient)
+	ghUser, _, err := ghClient.Users.Get(oauth2.NoContext, "")
+	if err != nil {
+		s.logger.Info(err)
+		return nil, internalError
```
sthaha, line 65: why is this an internal error if `Get` fails? shouldn't we look at what sort of an `err` it is before we say it is an `internal err`  or is it that the `err` is not nil only if there was a http issue?

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+		Scan(newUser).
+		Error; err != nil {
+		s.logger.Error(err)
+		return nil, internalError
+	}
+
+	return newUser, nil
+}
+
+func (s *service) createJWT(user *model.User) (string, error) {
```
sthaha, line 101: can we make this an independent  function and not a function of the `service` ?
the `s.logger` can be done by the caller.

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+var (
+	invalidCode   = auth.MakeInvalidCode(fmt.Errorf("invalid authorization code"))
+	internalError = auth.MakeInternalError(fmt.Errorf("failed to authenticate"))
+)
+
+// New returns the auth service implementation.
+func New(api app.Config) auth.Service {
+	return &service{api.Logger(), api.DB(), api.OauthConfig(), api.JWTSigningKey(), api.Data().Default.Scopes}
```
sthaha, line 47: we shouldn't Cache the DefaultScopes because the `Config.yaml` can change and the app.Data can get refreshed but the service will still use the old Scopes.

api/pkg/app/app.go
```diff
@@ -25,29 +25,48 @@ import (
...
 )
 
-// Config defines methods on APIConfig
-type Config interface {
+// Base defines methods on BaseConfig
+type Base interface {
 	Environment() EnvMode
 	Logger() *zap.SugaredLogger
 	DB() *gorm.DB
 	Data() *Data
```
sthaha, line 40: Can Data ever be nil?

api/pkg/app/app.go
```diff
@@ -25,29 +25,48 @@ import (
...
+	"golang.org/x/oauth2/github"
 
 	// Blank for package side effect: loads postgres drivers
 	_ "github.com/lib/pq"
 )
 
-// Config defines methods on APIConfig
-type Config interface {
+// Base defines methods on BaseConfig
+type Base interface {
```
sthaha, line 36: How about we can it `BaseConfig`

api/pkg/app/app.go
```diff
@@ -25,29 +25,48 @@ import (
...
 	Logger() *zap.SugaredLogger
 	DB() *gorm.DB
 	Data() *Data
 	Cleanup()
 }
 
-// APIConfig defines the configuration a services requires
-type APIConfig struct {
+// BaseConfig defines the base configuration every service requires
+type BaseConfig struct {
```
sthaha, line 45: How about we rename this to `APIBase` ?

api/pkg/app/app.go
```diff
@@ -84,49 +103,90 @@ func (db *Database) ConnectionString() string {
...
-	return ac.logger
+func (bc *BaseConfig) Logger() *zap.SugaredLogger {
+	return bc.logger
 }
 
 // Data returns Data object which consist data from config file
-func (ac *APIConfig) Data() *Data {
-	return &ac.data
+func (bc *BaseConfig) Data() *Data {
+	return &bc.data
```
sthaha, line 127: I think this should be return `Data` instead of `Data*`

api/pkg/app/app.go
SM43: Keeping the same for now

api/pkg/db/initializer/initiailizer.go
SM43: BaseConfig

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+	// gets access_token for user using authorization_code
+	token, err := s.conf.Exchange(oauth2.NoContext, p.Code)
+	if err != nil {
+		return nil, invalidCode
+	}
+
+	// gets user details from github using the access_token
+	oauthClient := s.conf.Client(oauth2.NoContext, token)
+	ghClient := github.NewClient(oauthClient)
+	ghUser, _, err := ghClient.Users.Get(oauth2.NoContext, "")
```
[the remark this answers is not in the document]
SM43: It is the `Response` object. It has GitHub API response.

api/pkg/service/auth/auth.go
SM43: The User is logged in when we get the `access_token`. So this will fail only when there is an HTTP issue. 
If we want we can user above `Response` Object to check status code. But I think this is kind of unexpected error so `internalError` would be fine. wdyt?

api/pkg/service/auth/auth.go
SM43: I moved that generic code to a package, created this map here because the user data is specific to this service.
We can also do `jwtToken.createForNormalUser` which will call `create` in that package.
WDYT? also what should be package name?

api/pkg/service/auth/auth_http_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/h2non/gock.v1"
+
+	"github.com/tektoncd/hub/api/gen/auth"
+	"github.com/tektoncd/hub/api/gen/http/auth/server"
+	"github.com/tektoncd/hub/api/pkg/testutils"
+)
+
+// Token for the user with github name "test-user" and github login "test"
+var ValidToken string = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9." +
```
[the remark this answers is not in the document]
SM43: It shouldn't be. updated

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+func (s *service) addUser(user *github.User) (*model.User, error) {
+
+	q := s.db.Model(&model.User{}).Where(&model.User{GithubLogin: user.GetLogin()})
+
+	newUser := &model.User{
+		GithubName:  user.GetName(),
+		GithubLogin: user.GetLogin(),
+	}
+	if err := q.FirstOrCreate(newUser).
+		Scan(newUser).
```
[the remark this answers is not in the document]
SM43: I used it for reading the `ID` after inserting but `FirstOrCreate` does that. I forgot to remove it. Thanks for pointing out..

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+	return newUser, nil
+}
+
+func (s *service) createJWT(user *model.User) (string, error) {
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+		"id":     user.ID,
+		"login":  user.GithubLogin,
+		"name":   user.GithubName,
+		"scopes": s.defaultScopes,
```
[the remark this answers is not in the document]
SM43: added a `data` field in `service` and which has a pointer to `Data`

api/pkg/service/auth/auth.go
SM43: Yeah added a `*Data` now.

api/cmd/db/main.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func main() {
-	api, err := app.FromEnv()
+	api, err := app.APIBaseFromEnv()
```
sthaha, line 26: This reads a bit odd .. but I don't have a good suggestion either :p

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/tektoncd/hub/api/pkg/app"
+	"github.com/tektoncd/hub/api/pkg/db/model"
+	jwtToken "github.com/tektoncd/hub/api/pkg/token"
+)
+
+type service struct {
+	logger        *zap.SugaredLogger
+	db            *gorm.DB
+	oauth         *oauth2.Config
+	data          *app.Data
```
sthaha, line 37: we shouldn't cache `*data` since this can change any time.

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+var (
+	invalidCode   = auth.MakeInvalidCode(fmt.Errorf("invalid authorization code"))
+	internalError = auth.MakeInternalError(fmt.Errorf("failed to authenticate"))
+)
+
+// New returns the auth service implementation.
+func New(api app.Config) auth.Service {
+	return &service{api.Logger(), api.DB(), api.OAuthConfig(), api.Data(), api.JWTSigningKey()}
```
sthaha, line 48: all access to Data and OAuthConfig can change anytime, so we shouldn't cache/store these.

api/pkg/service/auth/auth.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return nil, invalidCode
+	}
+
+	// gets user details from github using the access_token
+	oauthClient := s.oauth.Client(oauth2.NoContext, token)
+	ghClient := github.NewClient(oauthClient)
+	ghUser, _, err := ghClient.Users.Get(oauth2.NoContext, "")
+	if err != nil {
+		s.logger.Info(err)
```
sthaha, line 65: log.Error() instead of Info please

api/pkg/service/auth/auth.go
SM43: I agree with `Data`  but not sure with OAuth because we read it from env and we don't have re-read logic for it currently. I have changed the code for Data. But I do like the idea of re-reading OAuth env var. will need an API to invoke re-reading env var or config refresh API can take care of that.

PR #63, opened 2020-08-12
Api to update user's rating of a resource 

SM43:
Endpoint: PUT `/resource/<id>/rating`
id - ID of a resource
pass `rating` in the body
Use JWT of user in header Authorization to validate user for the API.

Issue #62 
# Submitter Checklist

review them:


api/design/rating.go
```diff
@@ -0,0 +1,53 @@
...
+
+var _ = Service("rating", func() {
+	Description("The rating service provides endpoint to read and write user's rating of resources")
+
+	Error("invalid-token", ErrorResult, "Invalid user token")
+	Error("invalid-scopes", ErrorResult, "Token scopes are invalid")
+	Error("internal-error", ErrorResult, "Internal server error")
+
+	Method("Get", func() {
+		Description("find user's rating of a resource")
```
PuneetPunamiya, line 29: Can we change it to `Find user's rating for a resource`

api/design/type.go
```diff
@@ -298,3 +298,9 @@ var Versions = ResultType("application/vnd.hub.versions", func() {
 
 	Required("latest", "versions")
 })
+
+var JWTAuth = JWTSecurity("jwt", func() {
+	Description("Secures endpoint by requiring a valid JWT token retrieved via the /auth/login endpoint.")
+	Scope("api:read", "Read-only access")
+	Scope("api:write", "Read and write access")
```
PuneetPunamiya, line 305: Do we need both scopes now ??

api/design/rating.go
```diff
@@ -0,0 +1,53 @@
...
+		Security(JWTAuth, func() {
+			Scope("api:read")
+		})
+		Payload(func() {
+			Attribute("id", UInt, "Id of a resource")
+			Token("token", String)
```
PuneetPunamiya, line 35: Can we add description for token ?

api/pkg/db/migration/202008121703_add_user_resource_rating_table.go
```diff
@@ -0,0 +1,42 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migration
+
+import (
+	"github.com/jinzhu/gorm"
+	"github.com/tektoncd/hub/api/pkg/db/model"
```
PuneetPunamiya, line 19: Can we please follow the same order to import ?

api/pkg/db/migration/202008121703_add_user_resource_rating_table.go
```diff
@@ -0,0 +1,42 @@
+// Copyright © 2020 The Tekton Authors.
```
PuneetPunamiya, line 1: Can we add this migration in `202006071000_add_resource_and_resource_versions_tables.go` file instead of adding a separate migration as it just creates the table for `userResourceRating` and adding foreign key constraints  which is done there as well???

api/pkg/db/model/model.go
```diff
@@ -76,4 +76,13 @@ type (
 		Login string `gorm:"not null;unique"`
 		Name  string `gorm:"not null;default:null"`
 	}
+
+	UserResourceRating struct {
```
PuneetPunamiya, line 80: Don't we need any not null constraints to any of the fields here ??

api/design/type.go
SM43: auth return jwt with both scopes so defining both at once would be okay I guess

api/pkg/db/migration/202008121703_add_user_resource_rating_table.go
SM43: Existing migration is never changed or updated. Migration runs only once. If you have an existing instance of DB when you run a migration, its ID is saved in the migration table, once ID is inserted that migration is not run again. So, adding anything in previous migration won't have any effect.

api/pkg/db/model/model.go
SM43: yeah rating can have a not null constraint

api/design/rating.go
```diff
@@ -50,4 +50,34 @@ var _ = Service("rating", func() {
...
 		})
 	})
+
+	Method("Update", func() {
+		Description("Update user's rating for a resource")
+		Security(JWTAuth, func() {
+			Scope("api:write")
+		})
+		Payload(func() {
+			Attribute("id", UInt, "ID of a resource")
```
PuneetPunamiya, line 60: Can we please add examples to the attributes ?

PuneetPunamiya:
@SM43  Thank you for the patch it overall looks good to me :slightly_smiling_face:

api/pkg/service/rating/rating.go
```diff
@@ -59,26 +60,91 @@ func (s *service) Get(ctx context.Context, p *rating.GetPayload) (*rating.GetRes
...
 	}
 
-	var res model.Resource
-	if err := s.db.First(&res, p.ID).Error; err != nil {
-		if gorm.IsRecordNotFoundError(err) {
-			return nil, notFoundError
-		}
-		s.logger.Error(err)
-		return nil, fetchError
+	_, err = s.checkIfResourceExists(p.ID)
```
sthaha, line 63: how about `validateResourceID()` ?

api/pkg/service/rating/rating.go
```diff
@@ -59,26 +60,91 @@ func (s *service) Get(ctx context.Context, p *rating.GetPayload) (*rating.GetRes
...
+	}
+
+	return s.updateResourceRating(res, p.ID)
+}
+
+func (s *service) updateUserRating(userID, resourceID, rating uint) error {
+
+	q := s.db.Where(&model.UserResourceRating{UserID: userID, ResourceID: resourceID})
+
+	rat := &model.UserResourceRating{}
```
sthaha, line 106: `rat` is a bit amusing and confusing :D .. how about calling it `r` or `rating` 😂

sthaha (commented):
/lgtm
/approve
/meow

PR #69, opened 2020-08-19
UI: Creates a store Categories using mobx-state-tree 

PuneetPunamiya:
# Changes

# Submitter Checklist

review them:


ui/public/config.js
```diff
@@ -0,0 +1,3 @@
+window.config = {
+    API_URL: 'https://api-tekton-hub-staging.apps.openshift-web.p0s5.p1.openshiftapps.com',
```
SM43, line 2: do we need to keep the stage URL in the code ?

ui/src/store/category.ts
```diff
@@ -0,0 +1,95 @@
...
+import { Api } from "../api";
+
+const Tag = types.model("Tags", {
+	id: types.integer,
+	name: types.string,
+});
+
+export const Category = types
+	.model("Category", {
+		id: types.number,
```
SM43, line 11: `types.number`  & `types.integer` ? which would be better to use? I guess `types.number` can hold float value also.

ui/src/api/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface Api {
+	categories(): Promise<ICategory>;
+}
+
+export class Hub implements Api {
+	async categories() {
+		try {
+			return axios.get(`${API_URL}/categories`);
+		} catch (err) {
+			return err.response;
```
SM43, line 13: What happens when the API server is down or takes a long time to respond?

SM43: Where would be the logic implemented for calling the API after some interval?

ui/src/api/index.ts
```diff
@@ -0,0 +1,16 @@
...
+		}
```
SM43, line 14: Where would be the logic implemented for calling the API after some interval?

ui/src/api/index.ts
PuneetPunamiya: So here we are handling and storing the error in the store whatever the error might and that error can you be used to display

ui/src/api/index.ts
PuneetPunamiya: Once we create the store then we can implement the logic there

ui/src/store/category.ts
PuneetPunamiya: While comparing the id while toggling the selected field of category we use id, but typescript don't have integer type whereas mst has both integer and number hence used number as type

vinamra28 (commented):
not a blocker, just some thoughts :sweat_smile:

ui/src/components/filter/Filter.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Checkbox } from "@patternfly/react-core"
+import { useObserver } from "mobx-react"
+
+interface store {
+  // TODO (puneet) -> modify type
+  store: any
+}
+
+const list = (props: store) => {
+  return props.store.map((c: any) => {
```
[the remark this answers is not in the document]
PuneetPunamiya: Yes I knew I could have given type as `ICategory` but since the filter component is generalized hence I was not sure what type it should have

ui/src/api/testutil.ts
```diff
@@ -0,0 +1,20 @@
...
+import { ICategory } from "../store/category"
+
+export class FakeHub implements Api {
+  dataDir: string
+
+  constructor(dataDir: string) {
+    this.dataDir = dataDir
+  }
+
+  categories() {
```
vinamra28, line 12: I guess this should be `async`

vinamra28:
I see there is an obsolete `App.tsx` file outside components dir ref:- https://github.com/PuneetPunamiya/hub/blob/category-mst/ui/src/App.tsx can we remove that if it's not used anywhere?

ui/.eslintrc.json
```diff
@@ -13,7 +13,8 @@
     "jest": true
   },
   "rules": {
-    "no-console": "warn"
+    "no-console": "warn",
+    "@typescript-eslint/explicit-module-boundary-types": "off"
```
pratap0007, line 17: can we define return type to exported function instead of defining rule here ?I guess will work.

vinamra28 (approved):
/lgtm

vinamra28 (approved):
/lgtm

ui/src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { Checkbox } from '@patternfly/react-core';
+import { useObserver } from 'mobx-react';
+import { Text, TextVariants } from '@patternfly/react-core';
+import { Grid, GridItem } from '@patternfly/react-core';
+import { Button } from '@patternfly/react-core/dist/js/components';
+import TimesIcon from '@patternfly/react-icons/dist/js/icons/times-icon';
```
karthikjeeyar, line 7:
```suggestion
import { Button, Checkbox, Text, TextVariants, Grid, GridItem } from '@patternfly/react-core';
import { useObserver } from 'mobx-react';
import { TimesIcon } from '@patternfly/react-icons';
```

ui/src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,65 @@
...
+import '@patternfly/react-core/dist/styles/base.css';
```
karthikjeeyar, line 8: I would suggest to move this to the top level of this app, so it can be applied everywhere instead of using the base css in the component files directly.

sthaha (commented):
/lgtm
/approve

PR #76, opened 2020-09-03
Updates Developement doc for authentication setup

SM43:
This updates the doc for setting up OAuth and how to create a jwt for testing.

# Submitter Checklist

review them:

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -1,6 +1,17 @@
-# Development <!-- TOC omit:true -->
+# Development 
 
-[toc]
+- [Getting Started](#getting-started)
+- [Checkout repository](#checkout-your-fork) 
+- [Requirements](#requirements)
+- [API Service](#api-service)
+  * [Running Database](#running-database)
+  * [Running Database Migration](#database-migration)
+  * [Adding GitHub OAuth Configuration](#adding-github-oAuth-configuration)
+  * [Running API Service](#running-api-service)
+  * [Running Tests](#running-tests) 
+  * [Creating JWT for testing](#creating-jwt-for-testing)
+  * [API Documentation](#api-documentation)
+
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


docs/DEVELOPMENT.md
```diff
@@ -112,124 +129,46 @@ To run the tests, we need a test db.
...
+````
+Add your OAuth Client ID from [.env.dev][env-dev] in HTML file in place of `<Add Client ID here>` and Save it.
 
+Open the HTML file, click on `Login with github` It will redirect you to GitHub. Once you authorize, it will redirect you to url localhost.
 
-#### Redeploy API Service
+for ex. `http://localhost:8080/?code=32d4a0b4eb6e9fbea731`
 
-**To redeploy api service after changes run the following command:**
+Use the `code` from url in `/auth/login` API. It will create a user in db and return a JWT.
```
PuneetPunamiya, line 163: Can we change this to 
By making a `POST` call to the `/auth/login` API, use the `code` from url in by passing it as params?

docs/DEVELOPMENT.md
```diff
@@ -86,6 +93,16 @@ Wait until the migration completes and logs to show
...
+
+Create a GitHub OAuth. You can find the steps to create it [here][gh-oauth]. 
+
+Use `http://localhost:8080` as the `Homepage URL` and `Authorization callback URL`. 
+
+After creation, add the OAuth Client ID as `GH_CLIENT_ID` and Client Secret as `GH_CLIENT_SECRET` in [.env.dev][env-dev].
+
+For `JWT_SIGNING_KEY`, you can use any random word.
+
 ### Running API Service
```
PuneetPunamiya, line 106: Here can we change the command to `go run ./cmd.api`

docs/DEVELOPMENT.md
```diff
@@ -112,124 +129,46 @@ To run the tests, we need a test db.
...
 ```
 
-This will expose the deployment as `LoadBalancer` service. You can access the api service using `LoadBalancer Ingress` and `Port`.
+**NOTE:** `tests` use the database configurations from [test/config/env.test][env-test-file]
 
-```bash
-kubectl describe svc api-load
-```
-Use the `LoadBalancerIngress:Port` eg. `http://35.223.169.99:8000`
+### Creating JWT for testing
```
PuneetPunamiya, line 142: Can we move this above tests ??

PuneetPunamiya:
@SM43  can we also add the the guide regarding agent api in this pr ??
cc @sthaha

docs/DEVELOPMENT.md
SM43: that would be okay I guess we are just letting the user know what he has to access. the user will have to look at the API definition :D before it is accessed

docs/DEVELOPMENT.md
SM43: :D

SM43:
> @SM43 can we also add the the guide regarding agent api in this pr ??
@PuneetPunamiya 
agent api will be used by only admins that too for creating agent for a cronjob or any other case we can document that in a doc..how to use API. but I think its not required in the development doc.

PuneetPunamiya:
@SM43  Thanks you
/lgtm

docs/DEVELOPMENT.md
```diff
@@ -86,149 +93,87 @@ Wait until the migration completes and logs to show
...
 ```
 
-The command above will create a migration image, push to registry pointed by `KO_DOCKER_REPO` and run a `job` that performs the migration.
-
-Check the logs using
-
+To re-generate the golden files use the below command
 ```bash
-kubectl logs job/db-migration
+go test ./pkg/... -test.update-golden
```
vinamra28, line 136: This test is failing on my machine :thinking: . Getting 
```
flag provided but not defined: -test.update-golden
```

SM43: command fails because some package doesn't have `testdata` dir as they don't use golden file for test.. so we need to run it on only those pkg which uses golden file.
```
go test $(go list -f '{{ .ImportPath }} {{ .TestImports }}' ./... | grep gotest.tools/v3/golden | awk '{print $1}' | tr '\n' ' ') -test.update-golden=true
```
this would fix that :smile:

vinamra28 (commented):
/lgtm

Issue #56, opened 2020-08-03
Use goa plugin for zap logger


SM43:
Use existing goa plugin for zap logger instead of creating a new one
[plugin](https://github.com/goadesign/plugins/tree/v3/zaplogger)

PR #80, opened 2020-09-07
Adds goa plugin for zap Suggared Logger

SM43:
Previously, we used to initialize zap logger but now replacing it
by using goa plugin which will initialize it.

Closes: #56

# Submitter Checklist

review them:


api/pkg/service/resource/resource.go
```diff
@@ -20,15 +20,15 @@ import (
...
 
 	"github.com/tektoncd/hub/api/gen/resource"
 	"github.com/tektoncd/hub/api/pkg/app"
 	"github.com/tektoncd/hub/api/pkg/db/model"
 )
 
 type service struct {
-	logger *zap.SugaredLogger
+	logger *log.Logger
 	db     *gorm.DB
```
sthaha, line 32: we can delete this db as well - right? 
None of the services would now need to have `db` - right?

SM43: I was yet to update the pr.. you reviewed early I guess :smile:

PuneetPunamiya:
@SM43  Thanks for the patch, the PR looks overall good to me :+1:

api/pkg/app/app.go
```diff
@@ -275,25 +288,16 @@ func initDB() (*Database, error) {
...
-
-	var log *zap.Logger
-	var err error
+func initLogger(mode EnvMode) *log.Logger {
 
 	switch mode {
 	case Production:
-		log, err = zap.NewProduction()
-
+		l, _ := zap.NewProduction()
```
sthaha, line 295: should we ignore the error here?

api/pkg/app/app.go
```diff
@@ -275,25 +288,16 @@ func initDB() (*Database, error) {
...
+		return &log.Logger{SugaredLogger: l.Sugar()}
 	default:
-		config := zap.NewDevelopmentConfig()
-		config.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
-		log, err = config.Build()
-	}
-
-	if err != nil {
-		return nil, err
+		l, _ := zap.NewDevelopment()
```
sthaha, line 298: isn't the earlier code better with `CapitalColorLevelEncoder`?

api/pkg/app/service.go
```diff
@@ -0,0 +1,67 @@
...
+// Service defines methods on BaseService
+type Service interface {
+	Logger(ctx context.Context) *log.Logger
+	DB(ctx context.Context) *gorm.DB
+}
+
+// BaseService defines configuraition for creating logger and
+// db object with http request id
+type BaseService struct {
+	name   string
```
sthaha, line 35: is the name used?

api/pkg/service/auth/service.go
```diff
@@ -39,16 +42,22 @@ var (
...
+func NewService(api app.Config, name string) *Service {
+	return &Service{
+		Service:    api.Service(name),
+		signingKey: api.JWTSigningKey(),
+	}
 }
 
 // JWTAuth implements the authorization logic for services for the "jwt" security scheme.
-func (v *Validator) JWTAuth(ctx context.Context, jwt string, scheme *security.JWTScheme) (context.Context, error) {
+func (v *Service) JWTAuth(ctx context.Context, jwt string, scheme *security.JWTScheme) (context.Context, error) {
```
sthaha, line 58: `s *Service` since it is Validator -> Service

api/pkg/service/auth/service.go
```diff
@@ -66,16 +75,22 @@ func (v *Validator) JWTAuth(ctx context.Context, jwt string, scheme *security.JW
...
+func (v *Service) User(ctx context.Context) (*model.User, error) {
+
+	logger := v.Logger(ctx)
+	db := v.DB(ctx)
+
 	userID := UserID(ctx)
-	log := v.Logger.With("user-id", userID)
+	log := &log.Logger{
+		SugaredLogger: logger.With(zap.String("user-id", strconv.FormatUint(uint64(userID), 10))),
+	}
```
sthaha, line 88: I think we would need to override the method `With` in `log.Logger` so that  services can use `With` to add additional fields.

api/pkg/app/app.go
SM43: yeah changed to prev

sthaha (commented):
/lgtm
/approve
/meow

PR #81, opened 2020-09-08
Modifies query api for resources

PuneetPunamiya:
# Changes

# Submitter Checklist

review them:


api/design/resource.go
```diff
@@ -30,13 +30,14 @@ var _ = Service("resource", func() {
 			Attribute("name", String, "Name of resource", func() {
 				Default("")
 			})
-			Attribute("kind", String, "Kind of resource", func() {
-				Enum("task", "pipeline", "")
-				Default("")
-			})
+			Attribute("kinds", ArrayOf(String), "Type of resource")
```
SM43, line 33: could you update the `Description` for `kind`,`tags`,`exact` ?

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +360,37 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+}
+
+func filterByTags(t []string) func(db *gorm.DB) *gorm.DB {
+	if t == nil {
+		return noop
+	}
+
+	return func(db *gorm.DB) *gorm.DB {
+		t = lower(t)
+		query := db.Model(&model.Resource{}).Select("DISTINCT(resources.id)").
```
SM43, line 381: move this to separate func?

api/pkg/service/resource/resource_test.go
```diff
@@ -29,7 +29,7 @@ func TestQuery_DefaultLimit(t *testing.T) {
 	testutils.LoadFixtures(t, tc.FixturePath())
```
SM43, line 29: could you please also add HTTP test?

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +347,43 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+
+func filterByTags(t []string) func(db *gorm.DB) *gorm.DB {
+	if t == nil {
+		return noop
+	}
+
+	return func(db *gorm.DB) *gorm.DB {
+		t = lower(t)
+
+		resID := getResourceIDs(t, db)
```
SM43, line 369: maybe we should call this from `Query` if tags != nil then call

SM43: for `filterByTags` only pass array of IDs and return the scope

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +347,43 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+
+		return db.Where("id IN (?)", resID)
+	}
+}
+
+func getResourceIDs(t []string, db *gorm.DB) (resID []uint) {
+
+	query := db.Model(&model.Resource{}).Select("DISTINCT(resources.id)").
```
SM43, line 377: Is `Select("DISTINCT(resources.id)")` requires now if we are using pluck ?

api/pkg/service/resource/resource.go
```diff
@@ -380,6 +418,25 @@ func matchesName(name string) func(db *gorm.DB) *gorm.DB {
...
+	if name == "" {
+		return noop
+	}
+	if exact {
+		return filterByName(name)
+	}
+	return matchesName(name)
+}
+
+// This function lowercase the first letter
```
SM43, line 431: lowercase the first letter? or the whole string?

api/pkg/service/resource/resource_http_test.go
```diff
@@ -42,7 +42,24 @@ func TestQuery_Http(t *testing.T) {
 	tc := testutils.Setup(t)
 	testutils.LoadFixtures(t, tc.FixturePath())
 
-	QueryChecker(tc).Test(t, http.MethodGet, "/query?name=build&kind=pipeline&limit=1").Check().
+	QueryChecker(tc).Test(t, http.MethodGet, "/query?name=build&kinds=pipeline&limit=1").Check().
```
SM43, line 45: could you please add more tests?
- passing only tags
- pass only kinds
- try to cover each case

api/pkg/service/resource/resource.go
PuneetPunamiya: Using pluck it selects a single column from the table it does not select unique id's from the table

api/pkg/service/resource/resource.go
PuneetPunamiya: It lowers the whole string, will update the comment

api/pkg/service/resource/resource.go
```diff
@@ -48,10 +48,16 @@ func New(api app.BaseConfig) resource.Service {
 // Find resources based on name, kind or both
 func (s *service) Query(ctx context.Context, p *resource.QueryPayload) (res resource.ResourceCollection, err error) {
 
+	var resID []uint
+	if p.Tags != nil {
+		resID = getResourceIDs(p.Tags, s.db)
```
SM43, line 53: queryResourcesByTags ?

api/pkg/service/resource/resource.go
SM43: `Pluck("DISTINCT(resources.id)", &resID)` will this work?

PuneetPunamiya: Yes this will work !

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +352,41 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+
+func queryResourcesByTags(t []string, db *gorm.DB) (resID []uint) {
+	t = lower(t)
+
+	query := db.Model(&model.Resource{}).Select("resources.id").
+		Joins("JOIN resource_tags on resource_tags.resource_id = resources.id").
+		Joins("JOIN tags on tags.id = resource_tags.tag_id").
+		Where("tags.name in (?)", t)
+
+	query.Pluck("DISTINCT(resources.id)", &resID)
```
SM43, line 385: Can you handle error in this query?

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +352,41 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+	return func(db *gorm.DB) *gorm.DB {
+
+		return db.Where("id IN (?)", t)
+	}
+}
+
+func queryResourcesByTags(t []string, db *gorm.DB) (resID []uint) {
+	t = lower(t)
+
+	query := db.Model(&model.Resource{}).Select("resources.id").
```
SM43, line 380: do we need `Select("resources.id")` now?

api/pkg/service/resource/resource.go
```diff
@@ -380,6 +421,24 @@ func matchesName(name string) func(db *gorm.DB) *gorm.DB {
...
+		return noop
+	}
+	if exact {
+		return filterByName(name)
+	}
+	return matchesName(name)
+}
+
+// This function lowercase all the elements of an array
+func lower(t []string) []string {
```
SM43, line 435: can you put this below `noop`?

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +352,41 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+		return noop
+	}
+
+	return func(db *gorm.DB) *gorm.DB {
+
+		return db.Where("id IN (?)", t)
+	}
+}
+
+func queryResourcesByTags(t []string, db *gorm.DB) (resID []uint) {
```
SM43, line 377: should we move this above `resourcesForQuery` and make this method of service and if the query fails return a `fetchError`?

SM43:
@PuneetPunamiya maybe we should add support for `catalog` also as a query param? wdyt?

PuneetPunamiya:
> @PuneetPunamiya maybe we should add support for `catalog` also as a query param? wdyt?

Yes I agree but thought once we have multi-catalog support then we can add it 
cc @sthaha

api/design/resource.go
```diff
@@ -30,13 +30,14 @@ var _ = Service("resource", func() {
...
-			Attribute("kind", String, "Kind of resource", func() {
-				Enum("task", "pipeline", "")
-				Default("")
-			})
+			Attribute("kinds", ArrayOf(String), "Kinds of resource")
+			Attribute("tags", ArrayOf(String), "Tags associated with a resource")
 			Attribute("limit", UInt, "Maximum number of resources to be returned", func() {
 				Default(100)
 			})
+			Attribute("exact", Boolean, "Find the exact resource", func() {
```
SM43, line 38: you can use an `Enum` here I guess to restrict the input...`true` and `false`

PuneetPunamiya: Since it is boolean do we still need to use enum ?

SM43: if we pass anything other that, does it return an error?

SM43: then if is fine

PuneetPunamiya: Yes that would return an error

PuneetPunamiya: Saying the type of exact should be boolean

api/pkg/service/resource/resource.go
```diff
@@ -158,6 +166,23 @@ func (s *service) ByID(ctx context.Context, p *resource.ByIDPayload) (res *resou
 	return res, nil
 }
 
+func (s *service) queryResourceIDByTags(t []string) ([]uint, error) {
+	t = lower(t)
```
SM43, line 170: should we rename `lower` to something else? :thinking: this is kinda odd

api/pkg/service/resource/resource.go
```diff
@@ -346,6 +371,28 @@ func filterByKind(t string) func(db *gorm.DB) *gorm.DB {
...
+		return noop
+	}
+
+	return func(db *gorm.DB) *gorm.DB {
+		t = lower(t)
+		return db.Where("LOWER(kind) IN (?)", t)
+	}
+}
+
+func filterByTags(t []uint) func(db *gorm.DB) *gorm.DB {
```
SM43, line 385: this is actually filter by id, we are passing resource ids... should we rename it to `filterByIds`

api/pkg/service/resource/resource.go
```diff
@@ -48,10 +48,18 @@ func New(api app.BaseConfig) resource.Service {
 // Find resources based on name, kind or both
 func (s *service) Query(ctx context.Context, p *resource.QueryPayload) (res resource.ResourceCollection, err error) {
 
+	var resID []uint
+	if p.Tags != nil {
+		if resID, err = s.queryResourceIDByTags(p.Tags); err != nil {
```
SM43, line 53: this can be filterByTags :D

sthaha (commented):
/lgtm
/approve
/meow

PR #87, opened 2020-09-19
Modifies status api to add db transaction

PuneetPunamiya:

# Changes

# Submitter Checklist

review them:


api/pkg/service/status/status_http_test.go
```diff
@@ -24,15 +24,18 @@ import (
...
 func TestOk_http(t *testing.T) {
+	tc := testutils.Setup(t)
+	testutils.LoadFixtures(t, tc.FixturePath())
 
 	checker := goahttpcheck.New()
 	checker.Mount(
 		server.NewStatusHandler,
 		server.MountStatusHandler,
-		status.NewStatusEndpoint(New()),
+		status.NewStatusEndpoint(New(tc)),
```
sthaha, line 38: could we please add a test where the `DB().Ping()` fails?

api/pkg/service/status/status.go
```diff
@@ -18,19 +18,29 @@ import (
...
+	return &service{api.Service("status")}
 }
 
 // Return status 'ok' when the server has started successfully
 func (s *service) Status(ctx context.Context) (res *status.StatusResult, err error) {
 
+	db := s.DB(ctx).DB()
+	log := s.Logger(ctx)
+
+	if err := db.Ping(); err != nil {
```
sthaha, line 40: it would be better if the json response contained the following format 

```yaml 
services:
  - name: api
     status: ok
 - name: db
    status: error
    error: "unable to reach db ... "
```

api/pkg/service/status/testdata/TestNotOk_http.golden
```diff
@@ -0,0 +1,13 @@
+{
+	"services": [
+		{
+			"name": "api",
+			"status": "ok"
+		},
+		{
+			"name": "db",
```
SM43, line 8: both negative test?

api/design/status.go
```diff
@@ -23,12 +23,7 @@ var _ = Service("status", func() {
 
 	Method("Status", func() {
 		Description("Return status 'ok' when the server has started successfully")
-		Result(func() {
-			Attribute("status", String, "Status of server", func() {
-				Example("status", "ok")
-			})
-			Required("status")
-		})
+		Result((server))
```
SM43, line 26: you can direct use `ArrayOf(Service)`... no need to define server I guess

api/design/type.go
```diff
@@ -282,3 +282,31 @@ var JWTAuth = JWTSecurity("jwt", func() {
 	Scope("rating:read", "Read-only access to rating")
 	Scope("rating:write", "Read and write access to rating")
 })
+
+var services = Type("services", func() {
```
SM43, line 286: service?

api/design/type.go
```diff
@@ -282,3 +282,31 @@ var JWTAuth = JWTSecurity("jwt", func() {
...
+	Description("Describes the services and their status")
+	Attribute("name", String, "Name of the service", func() {
+		Example("name", "api")
+	})
+	Attribute("status", String, "Status of the service", func() {
+		Example("status", "ok")
+	})
+	Attribute("error", String, "Internal server error", func() {
```
SM43, line 294: why `Internal server error` as description?

api/design/status.go
PuneetPunamiya: Yes I could use `ArrayOf(Service)` but that could give the response as 

```
[
         {
		"name": "api",
		"status": "ok"
	},
	{
		"name": "db",
		"status": "ok"
	}
]
```

and not as 

```
{
	"services": [
		{
			"name": "api",
			"status": "ok"
		},
		{
			"name": "db",
			"status": "ok"
		}
	]
}

```

api/design/type.go
PuneetPunamiya: If we use `service` it gives an error because it goa already has an interface with `Service` and the error is `Service is redeclared

api/design/status.go
SM43:
```
Attribute("services", ArrayOf("Service"), "Description")
```
won't this work inside Result?

api/design/type.go
SM43: Name it something else :smile:

api/pkg/service/status/status_http_test.go
```diff
@@ -15,24 +15,54 @@
...
+		assert.NoError(t, err)
+
+		golden.Assert(t, res, fmt.Sprintf("%s.golden", t.Name()))
+	})
+}
+
+func TestNotOk_http(t *testing.T) {
+	tc := testutils.Setup(t)
+
+	tc.DB().Close()
```
SM43, line 59: Do you know if the order of execution for the test is always the same? :thinking:
all the test in the package uses the same DB connection if `TestNotOk_http` executed first then the other will fail

SM43: can you check if the order of execution is always the same ?

sthaha: I don't think we should make assumptions about the order of tests.

SM43: then I think `tc.DB().Close()` is not a good idea because we use one db connection for all tests in a package ...wdyt?

api/design/type.go
SM43: `services` -> `server` ?

api/design/status.go
```diff
@@ -24,10 +24,7 @@ var _ = Service("status", func() {
 	Method("Status", func() {
 		Description("Return status 'ok' when the server has started successfully")
```
SM43, line 25: could you update the desc?

api/design/type.go
```diff
@@ -282,3 +282,18 @@ var JWTAuth = JWTSecurity("jwt", func() {
 	Scope("rating:read", "Read-only access to rating")
 	Scope("rating:write", "Read and write access to rating")
 })
+
+var server = Type("server", func() {
```
SM43, line 286: Server?

api/pkg/service/status/status.go
```diff
@@ -18,21 +18,37 @@ import (
...
 
 // New returns the status service implementation.
-func New() status.Service {
-	return &service{}
+func New(api app.BaseConfig) status.Service {
+	return &service{api.Service("status")}
 }
 
 // Return status 'ok' when the server has started successfully
 func (s *service) Status(ctx context.Context) (res *status.StatusResult, err error) {
```
SM43, line 35: ` ( *status.StatusResult,  error)`

api/pkg/service/status/status.go
```diff
@@ -18,21 +18,37 @@ import (
...
 
-	res = &status.StatusResult{
-		Status: "ok",
+	server := []*status.Server{}
+	server = append(server, &status.Server{Name: "api", Status: "ok"})
+
+	db := s.DB(ctx).DB()
+
+	res = &status.StatusResult{}
```
SM43, line 42: initialized at last also

api/pkg/service/status/status.go
```diff
@@ -18,21 +18,37 @@ import (
...
+	if err := db.Ping(); err != nil {
+		e := err.Error()
+		server = append(server, &status.Server{Name: "db", Status: "Error", Error: &e})
+		res = &status.StatusResult{Services: server}
+		return res, nil
```
SM43, line 47: `return &status.StatusResult{Services: server}, nil`?

api/pkg/service/status/status.go
```diff
@@ -18,21 +18,37 @@ import (
...
+	return &service{api.Service("status")}
 }
 
 // Return status 'ok' when the server has started successfully
 func (s *service) Status(ctx context.Context) (res *status.StatusResult, err error) {
 
-	res = &status.StatusResult{
-		Status: "ok",
+	server := []*status.Server{}
+	server = append(server, &status.Server{Name: "api", Status: "ok"})
```
SM43, line 38:
```
service := []*status.Services{
		&status.Services{
			Name: "api", Status: "ok",
		},
	}
```

api/pkg/service/status/status.go
```diff
@@ -18,21 +18,37 @@ import (
...
+		e := err.Error()
+		server = append(server, &status.Server{Name: "db", Status: "Error", Error: &e})
+		res = &status.StatusResult{Services: server}
+		return res, nil
 	}
+
+	server = append(server, &status.Server{Name: "db", Status: "ok"})
+
+	res = &status.StatusResult{Services: server}
 	return res, nil
```
SM43, line 53: `return &status.StatusResult{Services: server}, nil`?

sthaha (commented):
/lgtm
/approve
/meow

PR #89, opened 2020-09-22
CLI: tkn-hub 🐱

SM43:
- Initialize CLI for the hub (tkn-hub)
- Adds search command

```
 tkn-hub search <name> --kinds  --tags --match --output --limit
```
```
NAME            KIND       DESCRIPTION                                  TAGS
foo (0.1)       Task       Description for task abc version 0.1         tag3, tag1   
foo-bar (0.2)   Pipeline   Description for pipeline foo-bar versio...   ---  
```
# Submitter Checklist

review them:


cli/pkg/app/app.go
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"io"
+)
+
+const api string = "http://localhost:8000"
```
sthaha, line 21: should point to api.hub.tekton.dev. more importantly this should be overridable using an `ENV` or an `--api-server=https://my.hub.api.server` flag (viper? 🤔 )

cli/pkg/app/read.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		fmt.Fprintln(stream.Err, "Error: failed to read resource", err)
+		return nil, false, nil
+	}
+
+	switch status {
+	case 200:
+		return data, true, nil
+	case 404:
+		fmt.Fprintln(stream.Out, "No Resource Found")
```
sthaha, line 55: this should return an error and let the caller print or ignore instead of printing it here .. The rule this function breaks is SRP ... 
It not only fetches the resource (its responsibility)  but also prints it (another responsibility)

cli/pkg/app/read.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return data, resp.StatusCode, nil
+}
+
+// FetchHubResource will fetch resource from Hub and depending on status code
+// returns the data
+func FetchHubResource(stream *Stream, cli CLI, api string) ([]byte, bool, error) {
```
sthaha, line 42: Should't this be a method of cli itself?

cli/pkg/cmd/search/search_test.go
```diff
@@ -0,0 +1,198 @@
...
+		MinPipelinesVersion: "0.12",
+		RawURL:              "http://raw.github.url/foo-bar/",
+		WebURL:              "http://web.github.com/foo-bar/",
+		UpdatedAt:           "2020-01-01 12:00:00 +0000 UTC",
+	},
+	Tags: []*res.Tag{},
+}
+
+func TestSearch(t *testing.T) {
+	cli := test.New()
```
sthaha, line 84: test.NewCli ?

cli/pkg/formatter/field.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// FormatTags returns list of tags seperated by comma
+func FormatTags(tags []*client.TagResponse) string {
+
+	if len(tags) == 0 {
+		return "---"
+	}
+	var str string
+	for i, t := range tags {
```
sthaha, line 47: we should use a string.Builder instead of creating new string everytime

cli/pkg/formatter/field.go
```diff
@@ -0,0 +1,55 @@
...
+	if desc == "" {
+		return "---"
+	} else if len(desc) > 40 {
+		return desc[0:39] + "..."
+	}
+	return desc
+}
+
+// FormatTags returns list of tags seperated by comma
+func FormatTags(tags []*client.TagResponse) string {
```
sthaha, line 41: could we please write a unit-test for this 🤗 ?

cli/pkg/app/read.go
```diff
@@ -0,0 +1,64 @@
...
+
+	url := cli.API() + api
+	data, status, err := httpRead(url)
+	if err != nil {
+		fmt.Fprintln(stream.Err, "Error: failed to read resource", err)
+		return nil, false, nil
+	}
+
+	switch status {
+	case 200:
```
sthaha, line 52: can we please use Enum/const defined in the http package instead 🙏 ?

cli/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/tektoncd/hub/cli
```
vdemeester, line 1: I think we could use the following approach (if it works)

```
module github.com/tektoncd/hub/cli

require (
    // […]
    github.com/tektoncd/hub/api v0.0.0
)

replace github.com/tektoncd/hub/api => ../api
```

cli/pkg/app/app.go
```diff
@@ -0,0 +1,66 @@
...
+
+type Stream struct {
+	Out io.Writer
+	Err io.Writer
+}
+
+//func FetchHubResource(cli CLI, api string) ([]byte, bool, error) {
```
vdemeester, line 29: should probably be deleted :stuck_out_tongue_closed_eyes:

cli/pkg/app/hub.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+type Result interface {
+	Raw() ([]byte, error)
+}
```
vdemeester, line 11: I don't see where this interface is used

cli/pkg/app/hub.go
```diff
@@ -0,0 +1,45 @@
...
+
+type Hub interface {
+	Search(opt SearchOption) SearchResult
+
+	GetURL(endpoint string) ([]byte, int, error)
+}
+
+type HubClient struct {
+	api string
```
vdemeester, line 20: `s/api/URL` ?

cli/pkg/app/read.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+// // FetchHubResource will fetch resource from Hub and depending on status code
```
vdemeester, line 17: Should probably be deleted ?

cli/pkg/cmd/search/search.go
```diff
@@ -0,0 +1,167 @@
...
+		Kinds: opts.kinds(),
+		Name:  opts.name(),
+	})
+
+	out := opts.cli.Stream().Out
+	if opts.Output == "json" {
+		return printer.New(out).JSON(result.Raw())
+	}
+
+	return opts.print(result.Typed())
```
vdemeester, line 148: Seeing that we already get `opts.cli.Stream().Out` in this function **and** `opts.print` only uses that from `opts), I think this could be something more like

```go
type, err := result.Typed()
// handle error
return printer.New(out).Tabbed(typed, tmpl)
```

cli/pkg/cmd/search/testdata/TestSearch.golden
```diff
@@ -0,0 +1,3 @@
+NAME            KIND       DESCRIPTION                                  TAGS
+foo (0.1)       Task       Description for task abc version 0.1         tag3, tag1   
+foo-bar (0.2)   Pipeline   Description for pipeline foo-bar versio...   ---
```
vdemeester, line 3: Should "versions" be on there own column ?

cli/pkg/app/app.go
SM43: Yeah :smile:  today Sunil restructured the code a bit so code is gonna change now :sweat_smile:

cli/pkg/cmd/search/testdata/TestSearch.golden
SM43: yeah initially we added a separate column for the version `LATEST VERSION`, it was taking too much space :smile: so we appended that to name. should we have a separate column? WDYT?

cli/pkg/cmd/root.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	cli.SetStream(cmd.OutOrStdout(), cmd.OutOrStderr())
+
+	cmd.AddCommand(
+		search.Command(cli),
+	)
+
+	cmd.PersistentFlags().StringVar(
+		&apiURL, "api-server", "https://api.hub.tekton.dev", "Hub API Server URL")
```
sthaha, line 46: how about: `cli.HubURL` as the default where the package `cli` has

```
const HubURL = "https://...."
```

cli/pkg/cmd/search/testdata/TestSearch.golden
sthaha: yes, the same reasoning as @SM43 .. does `name (ver)` hinder readabilty?

sthaha (commented):
/lgtm
/meow

I will wait for @vdemeester to approve the new member to 😻 family.

cli/pkg/flag/kind.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// IsValid validates if a resource kind is valid Tekton kind
+func (r Kind) IsValid() error {
+	switch r {
+	case Task, Pipeline:
+		return nil
+	}
+	return fmt.Errorf("invalid value %q set for option kinds. Valid options: [task, pipeline]", r)
+}
```
vdemeester, line 35: Any reason to not use `tektoncd/pipeline` types ? (except not depending on it, which… make sense to me :wink: )

test/presubmit-tests.sh
```diff
@@ -119,6 +132,12 @@ run_unit_tests() {
   (
     set -eu -o pipefail
 
+    cd $CLI_DIR
+    cli-build
```
vdemeester, line 136: Any reason for not having `cli-build` in `run_build_tests` ?

vdemeester (commented):
2 small comments, but it looks good :tada:

popcor255:
This is so cool! I cant wait to use this! :smile_cat:

test/presubmit-tests.sh
sthaha: We should revise the whole of the CI script in a different PR. This just follows the pattern used for both api and ui.

cli/pkg/flag/kind.go
sthaha: +1 currently we don't need to depend on tektoncd/pipelines but in future we might have to esp.. if we want to support `tkn hub install task buildah` .

sthaha: @vdemeester is there a set of consts/enums that tektoncd/pipelines define we can reuse?

test/presubmit-tests.sh
sthaha: Although its outside the scope of this PR, I thought we should fix this for once and for all, hence the last commit that fixes it up. 
Thanks a lot for pointing this out @vdemeester .

vdemeester (approved):
/meow
:tada:

PR #105, opened 2020-10-06
Defines list of supported Tekton kind & add validation in Query API

SM43:
This defines list of supported Tekton kind by APIs. IsSupportedKind can be
used in CLI to check if a kind is supported by Hub APIs or not.

Also, adds validation for kinds in Query API.

# Submitter Checklist

review them:


api/pkg/parser/catalog.go
```diff
@@ -71,7 +72,7 @@ func (c *CatalogParser) Parse() ([]Resource, Result) {
 	resources := []Resource{}
 	result := Result{}
 
-	kinds := []string{"Task", "Pipeline"}
+	kinds := resource.SupportedKinds()
```
sthaha, line 75: I think we should keep this method in `parser` package, outside the services, so that services don't have to depend on each other.

SM43: /parser/resource/kind.go -> `resource.SupportedKinds()` ?

SM43: /pkg/resource/kind.go seems better to me than keeping in parser. wdyt?

sthaha (commented):
/lgtm
/approve
/meow

PR #111, opened 2020-10-13
Cleanup: removes addScopes from migration

SM43:
Currently, if we have to add a new scope, it will be added through
migration but from now on it can be added in config.yaml and if
the scope doesn't exist, it will be created while running the initializer.

# Submitter Checklist

review them:


sthaha (commented):
/lgtm
/approve

PR #112, opened 2020-10-13
Replaces hardcoded jwt with function to create it with required scopes

SM43:
This adds function to create jwt with required scopes on TestConfig
for tests instead of hardcoding jwt.

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/service/auth/auth_http_test.go
+++ b/api/pkg/service/auth/auth_http_test.go
@@ -23,24 +23,12 @@ import (
 
 	"github.com/ikawaha/goahttpcheck"
 	"github.com/stretchr/testify/assert"
-	goa "goa.design/goa/v3/pkg"
-	"gopkg.in/h2non/gock.v1"
-
 	"github.com/tektoncd/hub/api/gen/auth"
 	"github.com/tektoncd/hub/api/gen/http/auth/server"
 	"github.com/tektoncd/hub/api/pkg/testutils"
+	goa "goa.design/goa/v3/pkg"
+	"gopkg.in/h2non/gock.v1"
 )
-
-// Token for the user with github name "test-user" and github login "test"
-const validToken string = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9." +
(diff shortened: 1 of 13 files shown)


api/pkg/service/auth/auth_http_test.go
```diff
@@ -23,25 +23,13 @@ import (
 
 	"github.com/ikawaha/goahttpcheck"
 	"github.com/stretchr/testify/assert"
-	goa "goa.design/goa/v3/pkg"
-	"gopkg.in/h2non/gock.v1"
-
 	"github.com/tektoncd/hub/api/gen/auth"
 	"github.com/tektoncd/hub/api/gen/http/auth/server"
 	"github.com/tektoncd/hub/api/pkg/testutils"
+	goa "goa.design/goa/v3/pkg"
```
sthaha, line 29: is this part of `go fmt` ?

api/pkg/service/auth/auth_http_test.go
```diff
@@ -155,7 +147,11 @@ func TestLogin_Http_UserWithExtraScope(t *testing.T) {
 		marshallErr := json.Unmarshal([]byte(b), &res)
 		assert.NoError(t, marshallErr)
 
-		assert.Equal(t, validTokenWithExtraScope, res.Token)
+		// expected jwt for user
+		token, err := tc.UserWithScopes("foo", "foo:bar", []string{"rating:read", "rating:write", "agent:create"})
+		assert.NoError(t, err)
+
+		assert.Equal(t, token, res.Token)
```
sthaha, line 154: I think we must validate the `user.GitHubName` to be `foo`

api/pkg/service/auth/auth_http_test.go
SM43: yes we used to keep a line between 3rd party imports and internal imports but go fmt doesn't so that
so removing it in some file in this patch

SM43: gorm patch will update in remaining files

sthaha: we should rely on a single tool to do this and not mix it up. `go fmt` is the acceptable tool. Lets please stick to that 🙏 . Could you please revert this and get `go fmt` do its work ?

sthaha (commented):
Could we please remove changes unrelated to the patch and use `go fmt` for formatting?

sthaha (approved):
/lgtm
/approve
/meow

PR #113, opened 2020-10-13
Removes user validation in API methods

SM43:
As we are validating user while validating JWT so removing it from
API method

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/service/catalog/catalog.go
+++ b/api/pkg/service/catalog/catalog.go
@@ -55,11 +55,6 @@ func New(api app.Config) catalog.Service {
 // refresh the catalog for new resources
 func (s *service) Refresh(ctx context.Context, p *catalog.RefreshPayload) (*catalog.Job, error) {
 
-	user, err := s.User(ctx)
-	if err != nil {
-		return nil, err
-	}
-
 	log := s.Logger(ctx)
 	db := s.DB(ctx)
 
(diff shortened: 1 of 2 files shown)


api/pkg/service/catalog/catalog.go
```diff
@@ -70,7 +65,7 @@ func (s *service) Refresh(ctx context.Context, p *catalog.RefreshPayload) (*cata
 		return nil, notFoundOrInternalError(err)
 	}
 
-	job, err := s.wq.Enqueue(user.ID, ctg.ID)
+	job, err := s.wq.Enqueue(auth.UserID(ctx), ctg.ID)
```
sthaha, line 68: perhaps UserID() should be part of auth.Service itself so that the code reads `s.UserID(ctx)` just like `Logger(ctx)` . We should keep `auth.UserID` as well if we decide to go that direction.

sthaha (commented):
/lgtm
/approve
/meow

api/pkg/service/catalog/catalog.go
SM43: We can make `UserID` method of service.. then we can do `s.User()`

SM43: and this will be used only by API which needs authentication so the method of service would work. WDYT?

sthaha: I prefer having both :)

SM43: okay, will address it while adding expiry time in jwt. :smile_cat:

PR #115, opened 2020-10-14
Updates CatalogKindName API to return a resource instead of array

SM43:
This updates the API to return a single resource instead of an array
because in a catalog there won't be multiple resources of a kind
with the same name.

# Submitter Checklist

review them:


api/design/resource.go
```diff
@@ -161,47 +161,47 @@ var _ = Service("resource", func() {
...
+				Enum("task", "pipeline")
+			})
+			Attribute("name", String, "Name of resource", func() {
+				Example("name", "buildah")
+			})
+			Required("catalog", "kind", "name")
 		})
 		Result(Resource, func() {
-			View("default")
+			View("withoutVersion")
```
sthaha, line 200: we should return the versions: [...]

api/pkg/service/resource/resource.go
```diff
@@ -170,7 +170,7 @@ func (s *service) ByVersionID(ctx context.Context, p *resource.ByVersionIDPayloa
 }
 
 // find resources using name of catalog, resource name and kind of resource
-func (s *service) ByCatalogKindName(ctx context.Context, p *resource.ByCatalogKindNamePayload) (res resource.ResourceCollection, err error) {
+func (s *service) ByCatalogKindName(ctx context.Context, p *resource.ByCatalogKindNamePayload) (res *resource.Resource, err error) {
```
PuneetPunamiya, line 173: Do we need `res` here ? If m

PuneetPunamiya, line 173: Do we need `res` any more here ??

sthaha (commented):
/lgtm
/approve
/meow

SM43:
Recreated PR as #116

PR #116, opened 2020-10-19
Updates CatalogKindName API to return a resource instead of array

SM43:
This updates the CatalogKindName API to return a single resource instead of an array
because in a catalog there won't be multiple resources of a kind
with the same name.

# Submitter Checklist

review them:


sthaha (commented):
/lgtm
/approve
/meow

PR #123, opened 2020-10-22
Updates Response for APIs returning an array & the CLI commands using the APIs

SM43:
This updates the response of the Category & Resources APIs. Now, the object/array will be returned inside an object with a key `data` instead of only an object/array. so that later new fields can be added to the response
of an object without breaking the APIs.
for eg. metadata for pagination

Response:
```
{
   "data": {} | []
}

```

# Submitter Checklist

review them:


api/design/resource.go
```diff
@@ -73,13 +71,11 @@ var _ = Service("resource", func() {
...
 			Attribute("limit", UInt, "Maximum number of resources to be returned", func() {
-				Default(100)
+				Default(1000)
 				Example("limit", 100)
 			})
 		})
-		Result(CollectionOf(Resource), func() {
-			View("withoutVersion")
-		})
+		Result(ResourceResult)
```
PuneetPunamiya, line 78: Can't this be done like this
```
Result(func() {
     Attribute("data", CollectionOf(Resource), "List of resources", func() {
		View("withoutVersion")
	})
})
```

SM43: yeah it can be this way also.

PuneetPunamiya: So here we can avoid creating a new type in `type.go`

SM43: it is used for multiple APIs. and later when we add `metadata`, it has to be added at one place instead of multiple

SM43: thats why created a new type.

SM43: Will that be a issue?

PuneetPunamiya: Nope not any issue :+1:

sthaha (commented):
/hold

api/pkg/cli/hub/get_resource.go
```diff
@@ -39,6 +39,13 @@ type ResourceResult struct {
 	version string
 }
 
+// APIResponseForResource is the response of API when finding a resource
+type APIResponseForResource = rclient.ByCatalogKindNameResponseBody
```
sthaha, line 43: How about `ResourceResponse`?

api/pkg/cli/hub/get_resource.go
```diff
@@ -39,6 +39,13 @@ type ResourceResult struct {
...
+
+// APIResponseForResourceWithTheVersion is the response of API when finding a resource
+// with a specific version
+type APIResponseForResourceWithTheVersion = rclient.ByCatalogKindNameVersionResponseBody
```
sthaha, line 47: how about `ResourceWithVersionResponse` ?

sthaha, line 47: should we export these 🤔 ?

sthaha (commented):
/lgtm

We should wait for PR to experimental UI before merging this

/hold

PR #131, opened 2020-11-02
Upgrades gorm to 2.0

SM43:

# Submitter Checklist

review them:


api/pkg/db/initializer/initializer.go
```diff
@@ -149,10 +148,10 @@ func addUsers(db *gorm.DB, log *log.Logger, data *app.Data) error {
 			}
 
 			// Add scopes for user if not added already
-			us := model.UserScope{UserID: user.ID, ScopeID: scope.ID}
-			q = db.Model(&model.UserScope{}).Where(&us)
+			us := &model.UserScope{UserID: user.ID, ScopeID: scope.ID}
```
sthaha, line 151: could you please revert this to the old usage? 
```go
q = db.Model(...).Where(&us)
```
makes it quite clear the `us` is passed as a `reference` and may be modified.

api/pkg/service/auth/auth.go
```diff
@@ -20,14 +20,13 @@ import (
...
 	"github.com/google/go-github/github"
-	"github.com/jinzhu/gorm"
-	"golang.org/x/oauth2"
-
 	"github.com/tektoncd/hub/api/gen/auth"
 	"github.com/tektoncd/hub/api/gen/log"
 	"github.com/tektoncd/hub/api/pkg/app"
 	"github.com/tektoncd/hub/api/pkg/db/model"
 	"github.com/tektoncd/hub/api/pkg/token"
+	"golang.org/x/oauth2"
```
sthaha, line 28: is this how go fmt formatted this?

api/pkg/service/catalog/syncer.go
```diff
@@ -165,8 +165,8 @@ func (s *syncer) Process() error {
 
 	setJobState(model.JobRunning)
 
-	catalog := model.Catalog{}
-	db.Model(job).Related(&catalog)
+	catalog := &model.Catalog{}
+	db.Model(job).Association("Catalog").Find(catalog)
```
sthaha, line 169: could you please revert to the old style to pass `&catalog` to indicate that the object is modified?

api/pkg/service/catalog/syncer.go
```diff
@@ -146,15 +146,15 @@ func (s *syncer) Process() error {
 	log := s.logger.With("action", "process")
 	db := s.db
 
-	job := model.SyncJob{}
+	job := &model.SyncJob{}
 
 	// helper to update job state
 	setJobState := func(s model.JobState) {
 		job.SetState(s)
 		db.Model(&job).Updates(job)
```
sthaha, line 154: I think we should use `Model(&model.SyncJob{})` instead of `&job` since it is confusing to read.

api/pkg/service/catalog/syncer.go
```diff
@@ -195,26 +195,26 @@ func (s *syncer) Process() error {
...
 
-func (s *syncer) updateJob(job model.SyncJob, sha string, res []parser.Resource, result parser.Result) error {
+func (s *syncer) updateJob(job *model.SyncJob, sha string, res []parser.Resource, result parser.Result) error {
 	log := s.logger.With("action", "update-job", "job-id", job.ID)
 
 	txn := s.db.Begin()
 
-	catalog := model.Catalog{}
-	txn.Model(&job).Related(&catalog)
+	catalog := &model.Catalog{}
```
sthaha, line 203: same as above

api/pkg/service/resource/resource.go
```diff
@@ -471,3 +466,10 @@ func invalidKindError(kind string) error {
 	return resource.MakeInvalidKind(fmt.Errorf("resource kind '%s' not supported. Supported kinds are %v",
 		kind, parser.SupportedKinds()))
 }
+
+func isNotFoundError(err error) error {
```
sthaha, line 470: imho, this doesn't improve the readabilty. 
One would `isNotFoundError(err)` to return true if the `err` is `NotFound` and not a `notFoundError` or `fetchError`. 
Moreover, it fails to log.Error for all errors other than `NotFound`

api/pkg/service/status/status.go
```diff
@@ -38,7 +38,12 @@ func (s *service) Status(ctx context.Context) (*status.StatusResult, error) {
 		&status.HubService{Name: "api", Status: "ok"},
 	}
 
-	db := s.DB(ctx).DB()
+	db, err := s.DB(ctx).DB()
```
sthaha, line 41: isn't `api.DB().DB()` sufficient here?

api/pkg/service/auth/auth.go
SM43: yes

api/test/fixtures/resource_versions.yaml
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 - id: 1
-  version: 0.1
+  version: "0.1"
```
PuneetPunamiya, line 16: Is it the change in the gorm to change it to string or is it because we want to return it as a string in response???

sthaha: @PuneetPunamiya , yes it has to be a string to represent "0.1.1" or even "alpha" etc

sthaha (commented):
/lgtm

PR #136, opened 2020-11-18
API to refresh access token api for user

SM43:
This adds implementation for API to refresh access token for user. This
requires user refresh token to be passed to get a new access token.

# Submitter Checklist

review them:


api/design/auser.go
```diff
@@ -0,0 +1,57 @@
+// Copyright © 2020 The Tekton Authors.
```
PuneetPunamiya, line 1: Should it be 2021 ??

api/design/auser.go
```diff
@@ -0,0 +1,57 @@
...
+			Attribute("data", accessToken, "User Access JWT")
+			Required("data")
+		})
+
+		HTTP(func() {
+			POST("/user/refresh/accesstoken")
+			Header("refreshToken:Authorization")
+
+			Response(StatusOK)
+			Response("internal-error", StatusInternalServerError)
```
PuneetPunamiya, line 52: Are we missing invalid authorization code error ??

api/design/auser.go
SM43: yep

api/design/auser.go
SM43: why do we need `invalid authorization code error` ? this is not login api

PuneetPunamiya: Then why are we mentioning it above ??

api/pkg/service/user/user.go
```diff
@@ -0,0 +1,130 @@
+// Copyright © 2020 The Tekton Authors.
```
PuneetPunamiya, line 1: Same as above

api/design/auser.go
SM43: ohh.. willl remove it :P

api/design/auser.go
```diff
@@ -0,0 +1,57 @@
...
+package design
+
+import (
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("user", func() {
+	Description("The user service exposes endpoint to get user specific specs")
+
+	Error("invalid-code", ErrorResult, "Invalid Authorization code")
```
SM43, line 24: not needed

api/pkg/service/user/user.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	for _, s := range dbUser.Scopes {
+		userScopes = append(userScopes, s.Name)
+	}
+
+	return userScopes, nil
+}
+
+func createChecksum(token string) string {
```
PuneetPunamiya, line 127: Can you please add a unit test for this

api/design/user.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/tektoncd/hub/api/design/types"
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("user", func() {
+	Description("The user service exposes endpoint to get user specific specs")
+
+	Error("invalid-token", ErrorResult, "Invalid User token")
+	Error("invalid-scopes", ErrorResult, "Invalid User scope")
```
PuneetPunamiya, line 26: I don't think we have used this error in the implementation of the api

api/pkg/service/user/user.go
SM43: Do we need a separate test?  this is being tested with the API as we store it in db and check the db record.

api/design/user.go
SM43: It is being by the auth service while verifying jwt

api/pkg/service/user/user.go
PuneetPunamiya: In db we just check if we have checksum, what I think is passing the token it should return the checksum is what this function does and hence I think we need to test it

SM43: yeah but before checking in db it creates the checksum

PR #174, opened 2021-01-21
Refactors API Service for API versioning

SM43:

# Submitter Checklist

review them:


api/Dockerfile
```diff
@@ -11,8 +11,13 @@ RUN apk --no-cache add git ca-certificates && addgroup -S hub && adduser -S hub
 USER hub
 
 WORKDIR /app
+
 COPY --from=builder /go/src/github.com/tektoncd/hub/api/api-server /app/api-server
-COPY --from=builder /go/src/github.com/tektoncd/hub/api/gen/http/openapi3.yaml /app/gen/http/openapi3.yaml
+
+# For each new version, doc has to be copied 
+COPY gen/http/openapi3.yaml /app/gen/http/openapi3.yaml
+COPY v1/gen/http/openapi3.yaml /app/v1/gen/http/openapi3.yaml
```
PuneetPunamiya, line 19: So according to api policy will we make change here if further we have v2 and so on or keep this as well ??

api/pkg/shared/resource/resource.go
```diff
@@ -0,0 +1,332 @@
...
+	Catalog   string
+	VersionID uint
+}
+
+var (
+	FetchError    = fmt.Errorf("failed to fetch resources")
+	NotFoundError = fmt.Errorf("resource not found")
+)
+
+//Query resources based on name, kind, tags.
```
PuneetPunamiya, line 47: nit: Can we please add a space after `//`

api/v1/design/resource.go
```diff
@@ -0,0 +1,207 @@
...
+import (
+	types "github.com/tektoncd/hub/api/design/types"
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("resource", func() {
+	Description("The resource service provides details about all kind of resources")
+
+	HTTP(func() {
+		Path("/v1")
```
PuneetPunamiya, line 26: Just a thought is it possible to define the path somewhere so that we won't have to write the whole code again instead just take the path and rest all api design take it from generalized directory (this implies mostly when api's are not removed)

api/Dockerfile
SM43: for each new version, we need to copy the doc.. if we are deprecating a version then we will remove that doc

api/v1/design/resource.go
SM43: `define the path somewhere` where? 😄

PuneetPunamiya: What I meant was the same thing we have done by creating a `shared`  directory, not sure will that work or not just asking

SM43: I think for the design we can't do that we define the URL inside the service even if we share a service can't override any values

SM43: this is fine I guess easy to maintain

api/design/category.go
```diff
@@ -26,12 +27,11 @@ var _ = Service("category", func() {
 	Method("list", func() {
 		Description("List all categories along with their tags sorted by name")
 		Result(func() {
-			Attribute("data", ArrayOf(Category))
+			Attribute("data", ArrayOf(types.Category))
 		})
 
 		HTTP(func() {
 			GET("/categories")
-			GET("/v1/categories")
```
piyush-garg, line 34: This will be breaking change now as we already published

SM43: added back.. API will be removed in the next release and be internal to hub

piyush-garg:
/lgtm

Later, we may work on generalizing the API layer also like DB layer between multiple version services

PR #185, opened 2021-02-08
UI: Improves search

pratap0007:
This patch improves search with help of `debounce`
and adds searched input to the url as `search_query`

# Changes

# Submitter Checklist

review them:


ui/src/containers/Search/index.tsx
```diff
@@ -1,19 +1,39 @@
...
+import { UseDebounce } from './useDebounce';
 import './Search.css';
 
 const Search: React.FC = () => {
   const { resources } = useMst();
-
   const [value, setValue] = useState('');
+  let url = '';
+
+  // to get query param from thye url
```
PuneetPunamiya, line 13:
```suggestion
  // To get query param from the url
```

ui/src/containers/Search/index.tsx
```diff
@@ -1,19 +1,39 @@
...
+  const searchParams = new URLSearchParams(window.location.search);
```
PuneetPunamiya, line 14: What does this `window.location.search` do :thinking: ??

pratap0007: it finds query params from url

ui/src/containers/Search/index.tsx
```diff
@@ -1,19 +1,39 @@
...
 import { useHistory } from 'react-router-dom';
 import { useMst } from '../../store/root';
+import { UseDebounce } from './useDebounce';
 import './Search.css';
 
 const Search: React.FC = () => {
   const { resources } = useMst();
-
   const [value, setValue] = useState('');
+  let url = '';
```
vinamra28, line 11: since we are not using `url` anywhere except for line 31. Can we declare this there only?

ui/src/containers/Search/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react';
+import { useMst } from '../../store/root';
+
+export const UseDebounce = (value: string, delay: number) => {
+  const [debouncedValue, setDebouncedValue] = useState(value);
```
vinamra28, line 5: should we move this to a separate directory named utils where all the helper functions can be placed? :thinking: 
/cc @PuneetPunamiya

ui/src/containers/Search/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react';
+import { useMst } from '../../store/root';
+
+export const UseDebounce = (value: string, delay: number) => {
```
vinamra28, line 4: `UseDebounce` is a function and not a Functional Component right? Can we refractor this to `useDebounce`?

ui/package.json
```diff
@@ -56,15 +56,17 @@
     ]
   },
   "devDependencies": {
+    "@testing-library/react-hooks": "^5.0.3",
     "@types/react": "^16.9.49",
     "@typescript-eslint/eslint-plugin": "4.0.1",
     "@typescript-eslint/parser": "4.0.1",
+    "enzyme": "^3.11.0",
```
PuneetPunamiya, line 63: I see the version of the packages are similar then why are these changes done :thinking:

ui/src/containers/Search/Search.test.tsx
```diff
@@ -39,8 +45,10 @@ describe('search resources', () => {
         act(() => {
           node.props().onChange('golang');
         });
+        const setValue = jest.fn(() => 'golang');
 
-        expect(resources.search).toBe('golang');
+        expect(setValue()).toBe('golang');
+        expect(setValue).toHaveBeenCalledTimes(1);
```
PuneetPunamiya, line 51: Should we add a test once the value is cleared as well ??

ui/src/containers/Search/index.tsx
PuneetPunamiya: How does it gets cleared, are we storing it in localstorage and if someone's open with the query params will we get the expected result

ui/src/containers/Search/index.tsx
```diff
@@ -1,19 +1,38 @@
...
 
 const Search: React.FC = () => {
   const { resources } = useMst();
-
   const [value, setValue] = useState('');
 
-  const onSearchChange = (text: string) => {
-    setValue(text);
-    resources.setSearch(text.trim());
+  // to get query param from the url
```
PuneetPunamiya, line 12:
```suggestion
  // To get query params from the url
```

ui/src/utils/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react';
```
PuneetPunamiya, line 1: I think we need to add test for this as well

ui/src/containers/Search/index.tsx
```diff
@@ -1,19 +1,38 @@
...
+  const query = searchParams.get('search_query') || ' ';
+
+  useEffect(() => {
+    if (query !== ' ') {
+      resources.setSearch(query);
+      setValue(query);
+    }
+  }, [query, resources]);
+
+  const setParams = ({ query = '' }) => {
```
PuneetPunamiya, line 23: Not sure if this function should be here or in utils as well and also I think we need to add unit test for this function

ui/src/containers/KindFilter/__snapshots__/KindFilter.test.tsx.snap
```diff
@@ -19,8 +19,8 @@ exports[`KindFilter finds the filter component and matches the count 1`] = `
             </GridItem>
             <GridItem span={3}>
               <div className=\\"pf-l-grid__item pf-m-3-col\\">
-                <Button variant=\\"plain\\" aria-label=\\"Clear\\" onClick={[Function: res]}>
-                  <button onClick={[Function: res]} aria-disabled={false} aria-label=\\"Clear\\" className=\\"pf-c-button pf-m-plain\\" disabled={false} tabIndex={[undefined]} type=\\"button\\" data-ouia-component-type=\\"PF4/Button\\" data-ouia-safe={true} data-ouia-component-id={1}>
+                <Button variant=\\"plain\\" aria-label=\\"Clear\\" onClick={[Function: res] { _isMSTAction: true }}>
+                  <button onClick={[Function: res] { _isMSTAction: true }} aria-disabled={false} aria-label=\\"Clear\\" className=\\"pf-c-button pf-m-plain\\" disabled={false} tabIndex={[undefined]} type=\\"button\\" data-ouia-component-type=\\"PF4/Button\\" data-ouia-safe={true} data-ouia-component-id={1}>
```
vinamra28, line 23: I was wondering why the snapshot of the filters are changed? :thinking:

ui/src/containers/Search/index.tsx
```diff
@@ -1,19 +1,38 @@
...
+    const searchParams = new URLSearchParams();
+    searchParams.set('search_query', query);
+    return searchParams.toString();
   };
 
+  const updateURL = (text: string) => {
```
PuneetPunamiya, line 29: Same as above

ui/src/containers/Search/index.tsx
```diff
@@ -27,7 +46,11 @@ const Search: React.FC = () => {
     <TextInput
       value={value}
       type="search"
-      onChange={onSearchChange}
+      onChange={(text: string) => {
```
PuneetPunamiya, line 49:
```suggestion
      onChange={(resourceName: string) => {
```

PuneetPunamiya: Or may be `resource`

ui/src/containers/Search/index.tsx
vinamra28: @PuneetPunamiya I guess this is not a generic function which we can keep in utils and also for testing we'll have to export this function. I guess instead of this if it's possible can we test it's parent caller function?

ui/package.json
pratap0007: I guess these changes are done  by `npm i` to sort packages

ui/src/containers/KindFilter/__snapshots__/KindFilter.test.tsx.snap
pratap0007: I think bcoz of Search and App component changes,let me check again

pratap0007: these changes are happening bcoz of button's onclick function is a mst action

ui/package.json
PuneetPunamiya: I don't think it should, there's only one new package added

ui/src/containers/KindFilter/__snapshots__/KindFilter.test.tsx.snap
PuneetPunamiya: Are these changes expected ? Because I think all components are independent of each other

PuneetPunamiya: I think it might be because in filters we might be searching for a resource

ui/src/containers/Search/Search.test.tsx
```diff
@@ -39,8 +45,10 @@ describe('search resources', () => {
         act(() => {
           node.props().onChange('golang');
         });
+        const setValue = jest.fn(() => 'golang');
```
PuneetPunamiya, line 48: Should we call it `resourceName`

ui/src/containers/Search/index.tsx
PuneetPunamiya: @vinamra28  agreed on not keeping it in utils, but we can keep it in `common` dir as I think params will be required for all pages for e.g login and exporting a function won't be an issue

ui/src/utils/useDebounce.ts
PuneetPunamiya, line 1: The extension of the file should be with `.tsx` since we are using react hooks

ui/src/containers/Search/index.tsx
pratap0007: we are not storing it in localstorage , yeah will get the same result

ui/src/containers/Search/index.tsx
pratap0007: @vinamra28 @PuneetPunamiya while adding params to url for filters I will move it with the filters params  for now we can keep here , WDYT?

PR #199, opened 2021-02-23
Fixes yamllint and golangci lint errors

PuneetPunamiya:
# Changes

# Submitter Checklist

review them:


vdemeester (approved):

api/pkg/app/gorm_logger.go
```diff
@@ -69,7 +69,7 @@ func (w *prodWriter) Printf(format string, data ...interface{}) {
 				log = log.With("unknown", d)
 			}
 		default:
```
piyush-garg, line 71: can we remove default?

api/pkg/app/gorm_logger.go
```diff
@@ -69,7 +69,7 @@ func (w *prodWriter) Printf(format string, data ...interface{}) {
...
-			log = log.With("unknown", d)
+			_ = dt
```
SM43, line 72: `log = log.With("unknown", d)` any reason this was removed?

api/pkg/app/gorm_logger.go
SM43: why do we wanna remove it?

api/pkg/app/gorm_logger.go
PuneetPunamiya: Will add it

PuneetPunamiya: Removed by mistake while resolving the lint error

api/pkg/app/gorm_logger.go
piyush-garg: because we are doing nothing in default atm

api/pkg/app/gorm_logger.go
```diff
@@ -69,7 +69,7 @@ func (w *prodWriter) Printf(format string, data ...interface{}) {
 				log = log.With("unknown", d)
 			}
 		default:
-			log = log.With("unknown", d)
+			log = log.With("unknown", dt)
```
piyush-garg, line 72: Should not this be `d` ?

api/pkg/db/migration/migration_template.go
```diff
@@ -14,18 +14,21 @@
 
 package migration
 
-import (
-	"github.com/go-gormigrate/gormigrate/v2"
-	"github.com/tektoncd/hub/api/gen/log"
-	"gorm.io/gorm"
-)
+// Commenting this template as it is just a blueprint for adding a new migration and
```
piyush-garg, line 17: Should we completely remove this?

piyush-garg: or you can use `--skip-files`

api/pkg/app/gorm_logger.go
PuneetPunamiya: Then it would give a compile error of `dt` not used

SM43: i guess it should be `d` ..`dt` would be the type which is not required in the log

api/pkg/db/migration/migration_template.go
SM43: may be remove this.. we have a few migrations for examples now

PuneetPunamiya: Yes I think removing will be fine

api/pkg/app/gorm_logger.go
PuneetPunamiya: Would it be fine if we log as `log = log.With(dt, d)`

PuneetPunamiya: Since it is just a type

piyush-garg: You can use

```
switch d.(type) 
```

Then you don't require `_ = dt`

PuneetPunamiya: This would give `golangci-lint` error

api/pkg/app/gorm_logger.go
```diff
@@ -55,7 +55,8 @@ func (w *prodWriter) Printf(format string, data ...interface{}) {
 	msg := ""
 
 	for i, d := range data {
-		switch d.(type) {
+		dt := d
```
vinamra28, line 58: can't we remove this `dt` since we are using this only in the switch condition?

test/presubmit-tests.sh
```diff
@@ -122,18 +134,24 @@ run_unit_tests() {
...
+  )
 }
 
 run_integration_tests() {
   warn "No integration tests to run"
   return 0
 }
 
-main $@
+main $@
```
vinamra28, line 157: why this line is showing changes? :thinking:

api/pkg/app/gorm_logger.go
PuneetPunamiya: Added `dt` to resolve the lint errors

PR #210, opened 2021-03-05
Adds API policy doc

SM43:

# Submitter Checklist

review them:

### DIFF
--- a/docs/API_POLICY.md
+++ b/docs/API_POLICY.md
@@ -1,0 +1,65 @@
+# Hub API Policy
+
+This document proposes a policy regarding making updates to the APIs of this repo.
+
+NOTE: Hub Release Version and API version are different and may not be in sync. Making a Hub major release does not necessarily increment API version too. API version will be incremented only when there is a breaking change.
+eg. `/v1` -> `/v2` only when there is a breaking change.
+
+Hub API service hosts two types of APIs:
+
+## 1. Versioned APIs
+
+- These are the APIs which are exposed for external users to use and integrate with their applications.
+- These will have support provided by the Hub team.
+- These are the APIs which return hub resource data.
+- The version of the APIs will be prefixed to the API route. 
+  Eg. resource API - /v1/resources
+- Swagger Documentation for these APIs can be found at `https://api.hub.tekton.dev/<version>/schema/swagger.json`. Replace `<version>` with current version of APIs.
+ 
+
+#### Additive changes
+
+- Additive changes are changes that are added to the APIs in a backward compatible way and thus, do not cause problems for existing users of the API.
+- If the current version of API is `/v1` then any new addition for eg. adding new field in response of HTTP Request which will not affect existing users of v1 version of that API.
+- It is also possible to add new fields to request payload of an API as long as it is done in a backward compatible way, meaning it has a default value and can be omitted safely. 
+- These changes must be approved by at least 2 OWNERS.
+
+
+#### Breaking changes
+
+- Breaking changes are those when users have to change their implementation as per the changes in API. for eg. removal of a field from response, removing an API, removing a query param of an API. 
+- These changes will be introduced as a new version of the APIs.
+- If current version of APIs is `/v1` and an API is going to be removed then all other APIs will be moved to `/v2`. 
+  Users of v1 will be able to use old versioned APIs for at least 9 months then they will be removed.
+- These changes must be approved by at least 2 OWNERS.
+
+#### List of Versioned APIs
+  - `/v1`
+  - `/v1/query`
+  - `/v1/resources`
+  - `/v1/resource/<id>`
+  - `/v1/resource/<catalog>/<kind>/<name>`
+  - `/v1/resource/<catalog>/<kind>/<name>/<version>`
+  - `/v1/resource/<id>/versions`
+  - `/v1/resource/version/<versionID>`
+  - `/v1/schema/swagger.json`
+
+
+## 2. Internal APIs
+
+- These are the APIs which are specifically used for Hub UI and hub services.
+- These are not exposed for integration.
+- These may change at any time as the Hub UI evolves.
+- No support is provided for integration of this APIs.
+- The Swagger documentation can be found [here](https://api.hub.tekton.dev/schema/swagger.json).
+    Eg. Login, rating, Catalog refresh APIs, etc.
+
+#### List of Internal APIs
+- `/`
+- `/categories`
+- `/auth/login`
+- `/system/user/agent`
+- `/system/config/refresh`
+- `/catalog/refresh`
+- `/resource/<id>/rating`
+- `/schema/swagger.json`


vinamra28 (commented):
can we have removal section as well so that users can come to know that after how much time the APIs are going to be removed after being deprecated?

docs/API_POLICY.md
```diff
@@ -0,0 +1,30 @@
+# Hub API Policy
+This document proposes a policy regarding making API updates in this repo.
+
+There are 2 types of APIs:
+1. Versioned (For External Users): These are the APIs which are for users to use and will have support provided by the Hub team. These are the APIs which return tekton resources data. These are versioned APIs, the version of the APIs will be prefixed to the API route.
```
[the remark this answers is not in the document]
SM43: yup make sense will update

docs/API_POLICY.md
```diff
@@ -0,0 +1,30 @@
...
+#### Additive changes
+Additive changes are changes that add to the API and do not cause problems for users of previous versions of the API.
+These changes must be approved by at least 2 OWNERS.
+
+#### Backwards incompatible changes
+Backwards incompatible changes change the API, eg. removing a field in API response or removing an API.  These changes will mean that folks using a previous version of the API will need to adjust their usage in order to use the new version or an alternate API.
+
+#### Deprecated Features
+These are the features which will be removed in the next release but for folks using the previous  version would be able to use the feature for some amount of time.
+Eg. If a feature is deprecated from v2 to v3, then v2 APIs would be supported for x amount of time.
```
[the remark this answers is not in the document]
SM43: yeah that is for comments not sure what should be the time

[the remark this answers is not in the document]
SM43: maybe 9 months

docs/API_POLICY.md
PuneetPunamiya, line 5: Well the api's which are not versioned are also supported by Hub team right ??

PuneetPunamiya, line 5:
```suggestion
1. Versioned (For External Users): These are the APIs which are for users to use and will have support provided by the Hub team. These APIs return Hub resources data. The version of the APIs will be prefixed to the API route. 
```

docs/API_POLICY.md
```diff
@@ -0,0 +1,30 @@
+# Hub API Policy
+This document proposes a policy regarding making API updates in this repo.
+
+There are 2 types of APIs:
+1. Versioned (For External Users): These are the APIs which are for users to use and will have support provided by the Hub team. These are the APIs which return tekton resources data. These are versioned APIs, the version of the APIs will be prefixed to the API route. 
+Eg. resource API - /v1/resources
+2. Internal APIs: These are the APIs which are specifically used for Hub UI and hub services. These are not exposed for integration.
+Eg. Login, rating, Catalog refresh APIs, etc.
+
+### User facing APIs
```
PuneetPunamiya, line 10: What is the difference between this and the one which is mentioned above ???

PuneetPunamiya: i.e. the first type of api's `Versioned api's`

docs/API_POLICY.md
```diff
@@ -0,0 +1,30 @@
...
+- These APIs can be used for integration for users.
+- These APIs are versioned and have prefixed their version. Eg. /v1
+
+#### Additive changes
+Additive changes are changes that add to the API and do not cause problems for users of previous versions of the API.
```
PuneetPunamiya, line 15: Here when we specify owners, is it the Hub team ??

docs/API_POLICY.md
```diff
@@ -0,0 +1,30 @@
...
+These changes must be approved by at least 2 OWNERS.
```
PuneetPunamiya, line 16: Here when we specify owners, is it the Hub team ??

docs/API_POLICY.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Backwards incompatible changes
+Backwards incompatible changes change the API, eg. removing a field in API response or removing an API.  These changes will mean that folks using a previous version of the API will need to adjust their usage in order to use the new version or an alternate API.
```
PuneetPunamiya, line 19: Don't you think this should first follow the process of deprecation first and then removal, or is it that there can be some scenario where we have direct removal

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+This document proposes a policy regarding making API updates in this repo.
+
+There are 2 types of APIs:
+## 1. Versioned APIs: 
+- These are the APIs which are exposed for external users to use and integrate with their applications.
+- These will have support provided by the Hub team.
+- These are the APIs which return hub resource data.
+- These are versioned, the version of the APIs will be prefixed to the API route. 
+Eg. resource API - /v1/resources
+- Hub Release Version has nothing to do with API version. API version will be incremented only when there is a breaking change. Making a hub major release not necessarily increment API version too.
```
vinamra28, line 11: > Making a hub major release not necessarily increment API version too.

IMHO if there is a major release in API side then I guess we should be incrementing API version too as it will help users to differentiate between older and newer version. WDYT?

SM43: Hub is like a service that will keep evolving so the hub version doesn't really matter to the user as compared to any CLI version where the user needs to download the binary and run it.
Hub may have major changes in UI which might not have impact on API still incrementing API version would be unnecessary.
due to which hub release version and API version are independent. API version will increment only if there is a breaking change.

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
+# Hub API Policy
+This document proposes a policy regarding making API updates in this repo.
+
+There are 2 types of APIs:
+## 1. Versioned APIs: 
+- These are the APIs which are exposed for external users to use and integrate with their applications.
+- These will have support provided by the Hub team.
+- These are the APIs which return hub resource data.
+- These are versioned, the version of the APIs will be prefixed to the API route.
```
PuneetPunamiya, line 9:
```suggestion
The version of the APIs will be prefixed to the API route. 
```

docs/API_POLICY.md
PuneetPunamiya: So if I'm not wrong do you mean Hub release version and API release version might not go in sync in future ??

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
+# Hub API Policy
+This document proposes a policy regarding making API updates in this repo.
+
+There are 2 types of APIs:
+## 1. Versioned APIs: 
+- These are the APIs which are exposed for external users to use and integrate with their applications.
+- These will have support provided by the Hub team.
+- These are the APIs which return hub resource data.
+- These are versioned, the version of the APIs will be prefixed to the API route. 
+Eg. resource API - /v1/resources
```
PuneetPunamiya, line 10: I think we should also document somewhere which API's are versioned and which are not, may be here as it would help the user as well

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+- Hub Release Version has nothing to do with API version. API version will be incremented only when there is a breaking change. Making a hub major release not necessarily increment API version too.
+If hub is having a major release to v2.0.0 then it doesn't mean API will be incremented to `/v2`. 
+- Swagger Documentation for these APIs can be found at `https://api.hub.tekton.dev/<version>/schema/swagger.json`. Replace `<version>` with current version of APIs.
```
PuneetPunamiya, line 13: This would just return the `openAPI3,yaml` right  for that particular version ???

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+ 
+
+#### Additive changes
+- Additive changes are changes that add to the API and do not cause problems for exsiting users of the API.
+- If the current version of API is `/v1` then any new addition for eg. adding new field in response of a API will not affect existing users of v1 version of that API.
```
PuneetPunamiya, line 18: So in this case there will be an API release and not bumping the version of API right to /v2 ?

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+- These changes must be approved by at least 2 OWNERS.
+
+
+#### Breaking changes
+- Breaking changes are those when users have to change their implementation as per the changes in API. for eg. removal of a field from response, removing an API, removing a query param of an API.
```
PuneetPunamiya, line 23: Will there be any more cases ??

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+#### Additive changes
+- Additive changes are changes that add to the API and do not cause problems for exsiting users of the API.
+- If the current version of API is `/v1` then any new addition for eg. adding new field in response of a API will not affect existing users of v1 version of that API.
+- These changes must be approved by at least 2 OWNERS.
+
+
+#### Breaking changes
+- Breaking changes are those when users have to change their implementation as per the changes in API. for eg. removal of a field from response, removing an API, removing a query param of an API. 
+- This changes will be released as a new version of an API.
+- These changes will be introduced as a new version of the APIs.
```
PuneetPunamiya, line 25: Can you please list the difference between the above point and this point, I think both the points are same or is it that there will be Hub release which means the above point and the second says there's a version bump as well ??

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+- If current version of APIs is `/v1` and an API is going to be removed then all other APIs will be moved to `/v2`. Users of v1 will be able to use those APIs for 9 months then they will be removed.
```
PuneetPunamiya, line 26: I think we can say here that the API's will be deprecated first and then removed after 9 months wdyt ??

docs/API_POLICY.md
```diff
@@ -0,0 +1,39 @@
...
+- These changes must be approved by at least 2 OWNERS.
```
PuneetPunamiya, line 27: You mean Hub OWNERS ??

PuneetPunamiya (commented):
Thanks for this :call_me_hand: 
I just had one more question that is adding new field to the API will not bump the version of the API but removing any field would bump the version, why ? Because not sure if adding a new field might break the users usage as well

docs/API_POLICY.md
SM43: yes

docs/API_POLICY.md
SM43: yes later once the swagger is intergrated with UI we can just use that link

docs/API_POLICY.md
SM43: yes bumping the version to /v2 only if the API is breaking

docs/API_POLICY.md
SM43: not sure may be

docs/API_POLICY.md
SM43: yes

docs/API_POLICY.md
SM43: > ight break the us

yes adding a new field won't break the API, removing does.

SM43:
> Thanks for this 
> I just had one more question that is adding new field to the API will not bump the version of the API but removing any field would bump the version, why ? Because not sure if adding a new field might break the users usage as well

yes adding a new field won't break the API, removing does.

docs/API_POLICY.md
```diff
@@ -0,0 +1,64 @@
...
+- These will have support provided by the Hub team.
+- These are the APIs which return hub resource data.
+- The version of the APIs will be prefixed to the API route. 
+  Eg. resource API - /v1/resources
+- Swagger Documentation for these APIs can be found at `https://api.hub.tekton.dev/<version>/schema/swagger.json`. Replace `<version>` with current version of APIs.
+ 
+
+#### Additive changes
+
+- Additive changes are changes that are added to the APIs and do not cause problems for existing users of the API.
```
vdemeester, line 22: Additive changes are changes that are added to the APIs in a backward compatible way and thus, do not cause problems for existing users of the API.

docs/API_POLICY.md
```diff
@@ -0,0 +1,64 @@
...
+- If the current version of API is `/v1` then any new addition for eg. adding new field in response of a API will not affect existing users of v1 version of that API.
```
vdemeester, line 23: "in response" — you mean in the response of the HTTP request right ?
We should also state that adding a new field as part of the request payload is possible, as long as it is done in a backward compatible way, meaning it has a default value and can be omitted safely.

docs/API_POLICY.md
```diff
@@ -0,0 +1,64 @@
...
+- These changes must be approved by at least 2 OWNERS.
+
+
+#### Breaking changes
+
+- Breaking changes are those when users have to change their implementation as per the changes in API. for eg. removal of a field from response, removing an API, removing a query param of an API. 
+- These changes will be introduced as a new version of the APIs.
+- If current version of APIs is `/v1` and an API is going to be removed then all other APIs will be moved to `/v2`. 
+  Users of v1 will be able to use old versioned APIs for 9 months then they will be removed.
```
vdemeester, line 32: Are we saying that one a `v2` of an API is out, 9 months after, `v1` gets removed ?
(note : we should probably use "at least 9 month" 😝 )

docs/API_POLICY.md
```diff
@@ -0,0 +1,64 @@
...
+#### List of Versioned APIs
+  - `/v1`
+  - `/v1/query`
+  - `/v1/resources`
+  - `/v1/resource/<id>`
+  - `/v1/resource/<catalog>/<kind>/<name>`
+  - `/v1/resource/<catalog>/<kind>/<name>/<version>`
+  - `/v1/resource/<id>/versions`
+  - `/v1/resource/version/<versionID>`
+  - `/v1/schema/swagger.json`
```
vdemeester, line 44: Is this an API or just a file 😉

SM43: it is an API that returns a file on the server 😄

docs/API_POLICY.md
SM43: yup once `v2` is out `v1` will get removed after 9 months.

SM43: but let's say there is some breaking change in one of the API in `v1` so increment the version to `v2` so all other `v1` APIs would also be incremented or just the one?

SM43: my understanding was we increment all the APIs to v2 and then after 9 months remove all the v1 APIs

docs/API_POLICY.md
SM43: yup will add the 2nd point

docs/API_POLICY.md
vdemeester: yeah, it is probably better to bump the API version for all at the same time.

vdemeester (approved):

PR #213, opened 2021-03-15
Adds multi catalog support for UI

PuneetPunamiya:
- Initially `name` was used as an identifier to get the resource details
    which used to avoid adding resources with same name from different catalog.

 -  This patch add `uniqueKey` as an identifier which has value as
    `catalog-name/kind-name/resource-name` which adds support for
     muli-catalog

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/ui/src/components/Description/index.tsx
+++ b/ui/src/components/Description/index.tsx
@@ -6,10 +6,13 @@ import { Grid, Card, Tabs, Tab, GridItem, CardHeader, Spinner } from '@patternfl
 import { useMst } from '../../store/root';
 import Readme from '../Readme';
 import Yaml from '../Yaml';
+import { titleCase } from '../../common/titlecase';
 import './Description.css';
 
 interface Props {
   name: string;
+  catalog: string;
+  kind: string;
 }
 
 const Description: React.FC<Props> = (props) => {
@@ -20,7 +23,8 @@ const Description: React.FC<Props> = (props) => {
     setActiveTabKey(Number(tabIndex));
   };
 
-  const resource = resources.resources.get(props.name);
+  const { catalog, kind, name } = props;
+  const resource = resources.resources.get(`${catalog}/${titleCase(kind)}/${name}`);
 
   return useObserver(() =>
     resource.readme === '' || resource.yaml === '' ? (
--- a/ui/src/containers/BasicDetails/index.tsx
+++ b/ui/src/containers/BasicDetails/index.tsx
@@ -31,18 +31,20 @@ import { Icons } from '../../common/icons';
 import Icon from '../../components/Icon';
 import TooltipDisplay from '../../components/TooltipDisplay';
 import Rating from '../Rating';
+import { titleCase } from '../../common/titlecase';
 import './BasicDetails.css';
 
 const BasicDetails: React.FC = () => {
   const { resources } = useMst();
-  const { name } = useParams();
+  const { catalog, kind, name } = useParams();
 
-  const resource: IResource = resources.resources.get(name);
+  const resourceKey = `${catalog}/${titleCase(kind)}/${name}`;
+  const resource: IResource = resources.resources.get(resourceKey);
   const dropdownItems = resource.versions.map((value) => (
     <DropdownItem
       id={String(value.id)}
       key={value.id}
-      onClick={(e) => resources.setDisplayVersion(name, e.currentTarget.id)}
+      onClick={(e) => resources.setDisplayVersion(resourceKey, e.currentTarget.id)}
     >
       {value.version === resource.latestVersion.version
         ? `${value.version} (latest)`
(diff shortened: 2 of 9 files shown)


ui/src/components/Description/index.tsx
```diff
@@ -7,9 +7,12 @@ import { useMst } from '../../store/root';
 import Readme from '../Readme';
 import Yaml from '../Yaml';
 import './Description.css';
+import { titleCase } from '../../common/titlecase';
```
pratap0007, line 10: import re-order

ui/src/containers/Details/index.tsx
```diff
@@ -8,28 +8,22 @@ import Description from '../../components/Description';
...
+    const resourceKey = `${catalog}/${titleCase(kind)}/${name}`;
+    return resources.resources.has(resourceKey);
   };
 
   const resourceDetails = () => {
-    resources.versionInfo(name);
-    resources.loadReadme(name);
-    resources.loadYaml(name);
-    const resource = resources.resources.get(name);
+    const resourceKey = `${catalog}/${titleCase(kind)}/${name}`;
```
pratap0007, line 22: can we keep this globally ,since we are using same variable 2 times

ui/src/store/resource.test.ts
```diff
@@ -607,4 +609,42 @@ describe('Store functions', () => {
...
+          kind: 'Task',
+          latestVersion: 47,
+          displayVersion: 47,
+          tags: [1],
+          rating: 5,
+          versions: [47],
+          displayName: 'golang build'
+        });
+
+        store.add(item);
```
pratap0007, line 641: could you add a test to check resources size before adding a new resource

ui/src/store/resource.test.ts
```diff
@@ -607,4 +609,42 @@ describe('Store functions', () => {
...
+        expect(store.resources.size).toBe(8);
+
```
pratap0007, line 643: can we check catalogs size here as well ?

PuneetPunamiya: To add a catalog size we'll have to add a catalog first as well before

ui/src/containers/BasicDetails/index.tsx
```diff
@@ -32,17 +32,20 @@ import Icon from '../../components/Icon';
 import TooltipDisplay from '../../components/TooltipDisplay';
 import Rating from '../Rating';
 import './BasicDetails.css';
+import { titleCase } from '../../common/titlecase';
```
vinamra28, line 35: can we please re-order the imports ?

ui/src/store/resource.ts
```diff
@@ -41,7 +41,8 @@ const Version = types.model('Version', {
 export const Resource = types
   .model('Resource', {
     id: types.number,
-    name: types.identifier,
+    name: types.string,
+    uniqueKey: types.identifier,
```
vinamra28, line 45: instead of `uniqueKey` can we rename it to a better variable like `resourceKey`?

ui/src/store/resource.test.ts
```diff
@@ -607,4 +609,43 @@ describe('Store functions', () => {
...
+      () => {
+        const item = Resource.create({
+          id: 44,
+          name: 'golang-build',
+          uniqueKey: 'tekton-hub/Task/golang-build',
+          catalog: 2,
+          kind: 'Task',
+          latestVersion: 47,
+          displayVersion: 47,
+          tags: [1],
```
vinamra28, line 635: instead of creating a resource here shouldn't we update the `resource.json` with a new task having different catalog and test the left pane as well?

ui/src/containers/BasicDetails/index.tsx
```diff
@@ -32,17 +32,20 @@ import Icon from '../../components/Icon';
...
 
 const BasicDetails: React.FC = () => {
   const { resources } = useMst();
-  const { name } = useParams();
+  const { catalog, kind, name } = useParams();
 
-  const resource: IResource = resources.resources.get(name);
+  const resource: IResource = resources.resources.get(`${catalog}/${titleCase(kind)}/${name}`);
```
vinamra28, line 41: can we extract `${catalog}/${titleCase(kind)}/${name}` to a variable and reuse that variable?

vinamra28 (changes requested):

ui/src/store/resource.test.ts
PuneetPunamiya: Adding a resource in `resource.json` will change lot of test cases snapshot and also we won't be able to add a test to check if a resource is added from a different catalog

vinamra28: ohh...okay

vinamra28 (commented):
/lgtm
/meow

PR #215, opened 2021-03-18
API to refresh all catalogs

### LINKED ISSUE #207: Add an API to refresh all catalogs
Currently, we have an endpoint `/catalog/<catalogName>/refresh` where we pass the name of the catalog, check if the catalog exists, and refresh it.
But it would be good to have an API that will refresh all the catalog in DB.

Endpoint can be: `/catalog/refresh/all`
Users should have `catalog:refresh` scope to access the API

SM43:
This adds an endpoint /catalog/refresh/all which will refresh all
    catalogs in db. User will need to have catalog:refresh scope in token
    to access the API.
    
# Submitter Checklist

review them:

Closes #207


api/pkg/service/catalog/catalog.go
```diff
@@ -79,6 +79,35 @@ func (s *service) Refresh(ctx context.Context, p *catalog.RefreshPayload) (*cata
...
 }
 
+// refresh all catalog for new resources
+func (s *service) RefreshAll(ctx context.Context, p *catalog.RefreshAllPayload) ([]*catalog.Job, error) {
+
+	log := s.Logger(ctx)
+	db := s.DB(ctx)
+
+	ctgs := []model.Catalog{}
+	if err := db.Model(&model.Catalog{}).Find(&ctgs).Error; err != nil {
```
PuneetPunamiya, line 89: Can't we use directly `db.find` since we don't want any specific fields from the tables ?

api/design/catalog.go
```diff
@@ -26,18 +26,21 @@ var _ = Service("catalog", func() {
 	Error("not-found", ErrorResult, "Resource Not Found Error")
 
 	Method("Refresh", func() {
-		Description("Refreshes Tekton Catalog")
+		Description("Refresh a Tekton Catalog by its name")
```
PuneetPunamiya, line 29: I think since we are having this for multicatalog support it should be as 
`Refresh a Catalog by it's name`

piyush-garg:
If for a particular catalog we have

```
 /catalog/<catalog-name>/refresh
```

then for all catalogs, we can have

```
/catalog/refresh
```

It will also not break the old API. WDYT?

api/pkg/service/catalog/catalog.go
```diff
@@ -57,24 +57,28 @@ func (s *service) Refresh(ctx context.Context, p *catalog.RefreshPayload) (*cata
...
 	}
 
-	ret := &catalog.Job{ID: job.ID, Status: job.Status}
+	ret := &catalog.Job{ID: job.ID, CatalogName: ctg.Name, Status: job.Status}
 	log.Infof("job %d queued for refresh", job.ID)
 
 	return ret, nil
 }
+
+func catalogNotFoundErr(name string) error {
```
PuneetPunamiya, line 111: Why are we not adding this error in goa design and tweaking the error based on it ??

SM43:
> If for a particular catalog we have
> 
> ```
>  /catalog/<catalog-name>/refresh
> ```
> 
> then for all catalogs, we can have
> 
> ```
> /catalog/refresh
> ```
> 
> It will also not break the old API. WDYT?

yeah we can do that. to avoid the breakage `/catalog/refresh/all` which implies `all` catalog will be refreshed so added this 😄 
/catalog/refresh doesn't say it will refresh all catalogs. and as it is internal so didn't thought much of breakage

SM43:
API: /catalog/refresh

PR #217, opened 2021-03-18
Adds implementation and test for API: `/v1/catalogs`

PuneetPunamiya:
- This API will find and return an array of all catalogs
    with id, name and type as information of catalog

# Changes

# Submitter Checklist

review them:


api/pkg/service/catalog/catalog.go
```diff
@@ -78,3 +80,25 @@ func (s *service) Refresh(ctx context.Context, p *catalog.RefreshPayload) (*cata
...
+		log.Error(err)
+		return nil, internalError
+	}
+
+	res := &catalog.ListResult{
+		Data: []*catalog.Catalog{},
+	}
+
+	for _, c := range all {
+		res.Data = append(res.Data, &catalog.Catalog{ID: c.ID, Name: c.Name, Type: c.Type})
```
SM43, line 100: may be we should add url also?

SM43: if someone is integrating APIs then knowing which catalogs are in hub this API would be useful
I think we should expose it for users.. wdyt

piyush-garg: Yes, makes sense

SM43:
If we are going to support `/v1/catalogs` then we don't need `/catalogs`  hub UI can use the v1. so i think we should remove `/catalogs`.
wdyt?

api/v1/design/catalog.go
```diff
@@ -0,0 +1,48 @@
...
+
+var _ = Service("catalog", func() {
+	Description("The Catalog Service exposes endpoints to interact with catalogs")
+
+	HTTP(func() {
+		Path("/v1")
+	})
+
+	Error("internal-error", ErrorResult, "Internal Server Error")
+	Error("not-found", ErrorResult, "Resource Not Found Error")
```
SM43, line 30: is not-found required?

api/v1/design/catalog.go
```diff
@@ -0,0 +1,48 @@
...
+
+package design
+
+import (
+	types "github.com/tektoncd/hub/api/design/types"
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("catalog", func() {
+	Description("The Catalog Service exposes endpoints to interact with catalogs")
```
SM43, line 23: interact with catalogs? may be something else

PuneetPunamiya: Apologies will change the description

SM43: apologies 😄

api/v1/design/catalog.go
```diff
@@ -0,0 +1,46 @@
...
+
+package design
+
+import (
+	types "github.com/tektoncd/hub/api/design/types"
+	. "goa.design/goa/v3/dsl"
+)
+
+var _ = Service("catalog", func() {
+	Description("List of catalogs")
```
sm43, line 23: `The catalog service provides details about catalogs.`

sm43:
It seems there is a bug in goa, using `view` in the catalog generates code that fails the test in CLI.
the generated code in `gen/resource/service.go ` 
```
// NewViewedResources initializes viewed result type Resources from result type
// Resources using the given view.
func NewViewedResources(res *Resources, view string) *resourceviews.Resources {
	p := newResourcesView(res)
	return &resourceviews.Resources{Projected: p, View: "default"}
}
```
 this is used in cli to generated responses for API. but it si not considering the view and using the `default` view always
 could you please file a bug?

api/v1/service/catalog/catalog.go
```diff
@@ -0,0 +1,59 @@
...
+		log.Error(err)
+		return nil, internalError
+	}
+
+	res := &catalog.ListResult{
+		Data: []*catalog.Catalog{},
+	}
+
+	for _, c := range all {
+		res.Data = append(res.Data, &catalog.Catalog{ID: c.ID, Name: c.Name, Type: c.Type, URL: &c.URL})
```
pratap0007, line 55: could you please add `URL`  in required section of `type.go`

api/design/types/type.go
```diff
@@ -50,7 +50,7 @@ var Category = Type("Category", func() {
 	Required("id", "name", "tags")
 })
 
-var Catalog = Type("Catalog", func() {
+var Catalog = ResultType("Catalog", func() {
```
pratap0007, line 53:
```suggestion
var Catalog = ResultType("application/vnd.hub.catalog","Catalog", func() {
```

pratap0007, line 53:
```suggestion
var Catalog = ResultType("application/vnd.hub.catalog","Catalog", func() {
```

PR #219, opened 2021-03-22
Api to get resources from one or multiple catalogs

pratap0007:
This patch adds a catalogs params to query api and
which takes one or multiple catalogs name

endpoint: `/query?catalogs=<catalog_name>&catalogs=<catalog_name>`

# Submitter Checklist

review them:


api/v1/design/resource.go
```diff
@@ -33,12 +33,15 @@ var _ = Service("resource", func() {
...
 
 	Method("Query", func() {
-		Description("Find resources by a combination of name, kind and tags")
+		Description("Find resources by a combination of name, kind,catalog and tags")
 		Payload(func() {
 			Attribute("name", String, "Name of resource", func() {
 				Default("")
 				Example("name", "buildah")
 			})
+			Attribute("catalogs", ArrayOf(String), "Catalogs of resource to filter by", func() {
```
SM43, line 42: you need do this changes for non `/v1` (preview) API also

api/pkg/shared/resource/resource.go
```diff
@@ -314,6 +316,19 @@ func filterResourceName(match, name string) func(db *gorm.DB) *gorm.DB {
...
 
+func filterByCatalogs(catalogs []string) func(db *gorm.DB) *gorm.DB {
+	if len(catalogs) == 0 {
+		return noop
+	}
+
+	catalogs = lower(catalogs)
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Model(&model.Resource{}).
+			Joins("JOIN catalogs as ct on ct.id = resources.catalog_id").
```
SM43, line 327: I am not sure if we require `join` for this, we already `Preload` catalog data so we can just add a where statement. wdyt?

api/v1/design/resource.go
SM43: i said non-preview `also`

PuneetPunamiya:
@pratap0007  could you please rebase the pr once ??

api/pkg/shared/resource/resource_test.go
```diff
@@ -142,6 +142,26 @@ func TestQuery_ByTags(t *testing.T) {
...
+		Limit: 100,
+	}
+
+	res, err := req.Query()
+	assert.NoError(t, err)
+
+	assert.Equal(t, 1, len(res))
+	assert.Equal(t,"catalog-community",res[0].Catalog.Name)
+}
+
```
PuneetPunamiya, line 164: I think we should also add a negative test for this where we don't find any resource

PuneetPunamiya:
Thanks :call_me_hand: 

/lgtm

PR #221, opened 2021-03-23
UI: Updates Catalog store

pratap0007:
Previously we were fetching catalogs details from resources
in this patch we are fetching it from `/v1/catalogs` api
    
# Submitter Checklist

review them:

### DIFF
--- a/ui/public/config.js
+++ b/ui/public/config.js
@@ -1,4 +1,5 @@
 window.config = {
   API_URL: 'https://api.hub.tekton.dev',
-  GH_CLIENT_ID: ''
+  GH_CLIENT_ID: '',
+  API_VERSION: 'v1'
 };
--- a/ui/src/config/constants.tsx
+++ b/ui/src/config/constants.tsx
@@ -5,3 +5,4 @@ window.config = window.config || {
 
 export const API_URL = window.config.API_URL;
 export const GH_CLIENT_ID = window.config.GH_CLIENT_ID;
+export const API_VERSION = window.config.API_VERSION;
--- a/ui/src/store/catalog.test.ts
+++ b/ui/src/store/catalog.test.ts
@@ -1,7 +1,12 @@
-import { Catalog, CatalogStore } from './catalog';
 import { getSnapshot } from 'mobx-state-tree';
+import { when } from 'mobx';
+import { Catalog, CatalogStore } from './catalog';
 import { assert } from './utils';
 import { Icons } from '../common/icons';
+import { FakeHub } from '../api/testutil';
+
+const TESTDATA_DIR = `${__dirname}/testdata`;
+const api = new FakeHub(TESTDATA_DIR);
 
 describe('Store Object', () => {
   it('can create a catalog object', () => {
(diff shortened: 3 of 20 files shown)


ui/src/store/resource.ts
```diff
@@ -255,13 +258,15 @@ export const ResourceStore = types
...
 
         json.data.forEach((r: IResource) => {
-          self.catalogs.add(r.catalog);
           self.versions.put(r.latestVersion);
         });
 
         // adding the tags to the store - normalized
         const tags: ITag[] = json.data.flatMap((item: IResource) => item.tags);
 
+        const allcatalogs: ICatalog[] = json.data.flatMap((item: IResource) => item.catalog);
```
vinamra28, line 267: nit
```suggestion
        const allCatalogs: ICatalog[] = json.data.flatMap((item: IResource) => item.catalog);
```

ui/src/store/catalog.ts
```diff
@@ -73,4 +83,32 @@ export const CatalogStore = types
         .filter((c: ICatalog) => c.selected)
         .reduce((acc: string[], c: ICatalog) => [...acc, c.name], []);
     }
+  }))
+  .actions((self) => ({
```
vinamra28, line 87: should we add a space in between?

ui/src/config/constants.tsx
```diff
@@ -5,3 +5,4 @@ window.config = window.config || {
 
 export const API_URL = window.config.API_URL;
 export const GH_CLIENT_ID = window.config.GH_CLIENT_ID;
+export const API_VERSION = 'v1';
```
PuneetPunamiya, line 8: I think we should not hard code this, because once we have `v2` then again we'll have to change here, could you please read this through `window.config` as well ??

ui/src/store/root.tsx
```diff
@@ -36,7 +41,7 @@ const initRootStore = (api: Api) => {
     whitelist: ['accessTokenInfo', 'refreshTokenInfo', 'isAuthenticated', 'isLoading']
   });
 
-  return Root.create({}, { api, categories, resources, user });
+  return Root.create({}, { api, categories, resources, user, catalogs });
```
PuneetPunamiya, line 44:
```suggestion
  return Root.create({}, { api, catalogs, categories, resources, user });
```

ui/src/store/catalog.test.ts
```diff
@@ -15,62 +20,57 @@ describe('Store Object', () => {
...
+
+        store.clearSelected();
+
+        expect(catalogs.selected).toBe(false);
+        done();
+      }
+    );
   });
 
   it('can get an icon for catalog', (done) => {
```
PuneetPunamiya, line 76: Why are we not injecting `api` while creating the store for other tests ?

PuneetPunamiya (commented):
Please address the minor changes else pr looks fine to me

ui/src/config/constants.tsx
vinamra28: I guess we can keep this hard-coded so that we can make sure that UI is compatible whenever we have new version of API available otherwise there can be chances that UI might break

PuneetPunamiya: With adding it in `window.config` you can still get to know if it might break the ui or not

PuneetPunamiya: In short even if we get it through `window.config` we can make sure that whethere it is compatible with UI or not

ui/src/store/catalog.test.ts
pratap0007: not required, but needs to modify test

ui/public/config.js
```diff
@@ -1,4 +1,5 @@
 window.config = {
   API_URL: 'https://api.hub.tekton.dev',
-  GH_CLIENT_ID: ''
+  GH_CLIENT_ID: '',
+  API_VERSION: ''
```
PuneetPunamiya, line 4: Could you please update the config map and deployment manifest as well ??

PuneetPunamiya (commented):
I guess we'll also have to update the `start.sh` file by adding `API_VERSION` there

vinamra28 (commented):
/lgtm

ui/public/config.js
```diff
@@ -1,4 +1,5 @@
 window.config = {
   API_URL: 'https://api.hub.tekton.dev',
-  GH_CLIENT_ID: ''
+  GH_CLIENT_ID: '',
+  API_VERSION: 'v1'
```
PuneetPunamiya, line 4: I think we shouldn't add this

pratap0007: we should add this to keep `v1` as default ,I guess otherwise catalog wouldn't able to access

PuneetPunamiya:
Thanks :call_me_hand: 
/approve
/meow
/woof

PR #224, opened 2021-03-28
API to get res versions compatible with a tekton pipelines version

sm43:
Adds Endpoint `/resource/{catalog}/{kind}/{name}/pipelinesversion/{pipelinesVersion}`
which returns an array of resource version compatible with the pipelines version
passed.

# Submitter Checklist

review them:


api/.env.dev
```diff
@@ -6,11 +6,11 @@ POSTGRES_DB="hub"
 POSTGRES_USER="postgres"
 POSTGRES_PASSWORD="postgres"
 
-GH_CLIENT_ID=""
-GH_CLIENT_SECRET=""
+GH_CLIENT_ID="client-id"
```
PuneetPunamiya, line 9: Do we need this change ??

sm43: every time we need to run the server we have to add some values let some default values be there. its in dev so won't affect anything, we can add in development.md to change as per requirement

sm43:
based on discussion 
for now `/resource/<catalog>/<kind>/<name>` will be added a query param `minPipelineversion` and return the latest compatible version 
/close

PR #226, opened 2021-03-30
UI: Adds icon for catalogs

pratap0007:
# Submitter Checklist

review them:


ui/src/store/catalog.ts
```diff
@@ -2,9 +2,7 @@ import { Instance, types } from 'mobx-state-tree';
 import { Icons } from '../common/icons';
 
 const icons: { [catalog: string]: Icons } = {
-  official: Icons.Cat,
-  verified: Icons.Certificate,
-  community: Icons.User
+  catalog: Icons.Catalog
```
vinamra28, line 5: shouldn't we mark this as `community` instead of `catalog` since we are getting `community` from the API call?

ui/src/store/catalog.ts
```diff
@@ -21,7 +19,7 @@ export const Catalog = types
   }))
   .views((self) => ({
     get icon(): Icons {
-      return icons[self.type] || Icons.Unknown;
+      return icons[self.type] || icons.catalog;
```
vinamra28, line 22: shouldn't we keep this as `Icons.unknown`?

ui/src/store/catalog.ts
PuneetPunamiya: And also are we using `community` anywhere anymore ??

pratap0007: now there is only one icon for any type of catalog so kept it as `catalog`

ui/src/store/catalog.ts
vinamra28, line 22:
```suggestion
      return icons[self.type] || Icons.catalog;
```

ui/src/store/catalog.ts
vinamra28: I guess as of now we are adding new catalogs as community catalog right ?

pratap0007: yeah, I will update it

vinamra28 (commented):
/lgtm

PR #227, opened 2021-03-30
Minor bug fix in Hub CLI info command

PuneetPunamiya:
- Fixes the logic by adding a break statement
     in formatting the description of resource in
     Hub CLI info command

   - Adds tests for wrapText function which formats
     the description of resource

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/cli/formatter/field.go
+++ b/api/pkg/cli/formatter/field.go
@@ -108,6 +108,7 @@ func breakString(desc string, width, titleLength int) string {
 	for i := firstLineEnd; i < descLength; i = i + spaceIndex {
 		if descLength < i+width {
 			sb.WriteString(desc[i:])
+			break
 		} else {
 			spaceIndex = findSpaceIndexFromLast(desc[i : i+width])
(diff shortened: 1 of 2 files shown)


api/pkg/cli/formatter/field.go
```diff
@@ -108,6 +108,7 @@ func breakString(desc string, width, titleLength int) string {
 	for i := firstLineEnd; i < descLength; i = i + spaceIndex {
 		if descLength < i+width {
 			sb.WriteString(desc[i:])
+			break
```
piyush-garg, line 111: Why we need this? the condition in for look will do this

SM43: `i` is incremented by adding `spaceIndex` but in the first case we didn't need to calculate it as it is the last line so it adds `0` in `i` and it goes in an infinite loop, in else we calculate `spaceIndex` as we need to break the line at that index.

vinamra28 (commented):
/lgtm

PR #231, opened 2021-03-31
UI: Adds user informations after authentication

pratap0007:
This patch displays user informations such as
`github profile picture` and `user name`
after successfully authentication

Screenshot from 2021-03-31 22-21-05

# Submitter Checklist

review them:

### DIFF
--- a/ui/src/api/index.ts
+++ b/ui/src/api/index.ts
@@ -2,7 +2,7 @@ import axios from 'axios';
 import { API_URL } from '../config/constants';
 import { ICategory } from '../store/category';
 import { IResource, IVersion } from '../store/resource';
-import { ITokenInfo } from '../store/auth';
+import { ITokenInfo, IUserProfile } from '../store/auth';
 
 interface Token {
   token: string;
@@ -35,6 +35,7 @@ export interface Api {
   setRating(resourceId: number, token: string, rating: number): Promise<void | null>;
   getRefreshToken(refreshToken: string): Promise<ITokenInfo>;
   getAccessToken(accessToken: string): Promise<ITokenInfo>;
+  profile(token: string): Promise<IUserProfile>;
 }
 
 export class Hub implements Api {
@@ -162,6 +163,21 @@ export class Hub implements Api {
           Authorization: `Bearer ${refreshToken}`
         }
       })
+        .then((response) => response.data)
+        .catch((err) => Promise.reject(err.response));
+    } catch (err) {
+      return err;
+    }
+  }
+
+  async profile(token: string) {
+    try {
+      return axios
+        .get(`${API_URL}/user/info`, {
+          headers: {
+            Authorization: `Bearer ${token}`
+          }
+        })
         .then((response) => response.data)
         .catch((err) => Promise.reject(err.response));
     } catch (err) {
--- a/ui/src/store/auth.test.ts
+++ b/ui/src/store/auth.test.ts
@@ -166,4 +166,31 @@ describe('Store functions', () => {
       }
     );
   });
+
+  it('can get user information after logged in', (done) => {
+    const user = AuthStore.create(
+      { accessTokenInfo: {}, refreshTokenInfo: {}, profile: {} },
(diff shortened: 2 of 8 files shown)


ui/src/api/index.ts
```diff
@@ -35,6 +35,7 @@ export interface Api {
   setRating(resourceId: number, token: string, rating: number): Promise<void | null>;
   getRefreshToken(refreshToken: string): Promise<ITokenInfo>;
   getAccessToken(accessToken: string): Promise<ITokenInfo>;
+  info(token: string): Promise<IUserInfo>;
```
vinamra28, line 38: I guess we can rename this to userInfo. WDYT?

ui/src/store/auth.ts
```diff
@@ -193,6 +206,23 @@ export const AuthStore = types
         };
         self.setErrorMessage(error);
       }
+    }),
+
+    getInfo: flow(function* () {
```
vinamra28, line 211: same here. We can make it as getUserInfo

ui/src/store/auth.ts
```diff
@@ -29,6 +36,7 @@ export const AuthStore = types
   .model('AuthStore', {
     accessTokenInfo: types.optional(TokenInfo, {}),
     refreshTokenInfo: types.optional(TokenInfo, {}),
+    info: types.optional(UserInfo, {}),
```
vinamra28, line 39: same userInfo

ui/src/api/index.ts
pratap0007: we have `user` as AuthStore instnace in the root store so here I have used `info`. if I rename it to `userInfo` then `user.userInfo` might be a bit confusing , WDYT?

PuneetPunamiya: Can we rename it to `profile` which will call the function as `user.Profile`

ui/src/store/auth.ts
PuneetPunamiya: Are we not changing here to profile ?

PuneetPunamiya (commented):
I think if we are using `profile` then I guess we should add it everywhere in order to maintain consistency

ui/src/store/auth.ts
pratap0007: just missed :smile: will update

ui/src/store/auth.test.ts
```diff
@@ -166,4 +166,31 @@ describe('Store functions', () => {
       }
     );
   });
+
+  it('can get user information after logged in', (done) => {
+    const store = AuthStore.create(
```
PuneetPunamiya, line 171: Can we use the variable name `user` here instead of `store`

ui/src/store/auth.ts
```diff
@@ -46,6 +54,11 @@ export const AuthStore = types
       self.refreshTokenInfo.expiresAt = item.expiresAt;
       self.refreshTokenInfo.refreshInterval = item.refreshInterval;
     },
+    addUserInfo(item: IUserInfo) {
```
PuneetPunamiya, line 57:
```suggestion
    addUserProfile(item: IUserProfile) {
```

ui/src/store/auth.ts
```diff
@@ -8,12 +8,19 @@ export const TokenInfo = types.model('TokenInfo', {
...
+  avatarUrl: types.optional(types.string, '')
+});
+
 export const Error = types.model('Error', {
   status: types.optional(types.number, 0),
   customMessage: types.optional(types.string, ''),
   serverMessage: types.optional(types.string, '')
 });
 
+export type IUserInfo = Instance<typeof UserInfo>;
```
PuneetPunamiya, line 23:
```suggestion
export type IUserProfile = Instance<typeof UserInfo>;
```

PuneetPunamiya (commented):
Thanks :call_me_hand: 
/approve
/meow
/woof

vinamra28 (commented):
/lgtm

PR #232, opened 2021-04-05
Allows Task, Pipeline in title case to be used as kind in APIs

sm43:
Currently, when we use Task or Pipeline in APIs
- `resource/<catalog>/<kind>/<name>`
- `resource/<catalog>/<kind>/<name>/<version>`
it would return an error asking to use them in lower case.
This patch will allow to use it in the title case as well.

# Submitter Checklist

review them:


api/design/resource.go
```diff
@@ -121,7 +121,7 @@ var _ = Service("resource", func() {
 				Example("catalog", "tektoncd")
 			})
 			Attribute("kind", String, "kind of resource", func() {
-				Enum("task", "pipeline")
+				Enum("task", "pipeline", "Task", "Pipeline")
```
sthaha, line 124: I think hub api service should only allow `lower case`. It should be responsibility of the cli tool / client tools to convert from `Pipeline` to `pipeline`

sm43: @sthaha 
Case 1 :
`/resource/Tekton/Task/Buildah` returns an error because `Task` should be `task`.  catalog name and resource name can be in title case
Case 2:
all our API responses have `kind`: Task Pipeline in the title case but API doesn't allow input kind as Title case.
due to which decided to allow title case. WDYT?
should the response be in lower case?

sm43:
based on discussion, no changes would be done in API and the client can handle it before calling
/close

PR #235, opened 2021-04-06
UI: Changes navigation url from home page to previous page 

pratap0007:
-    Changes navigation url from home page to previous page url
      - Previously navigation url was home page after successful authentication now it will be previous page url
      - Removes force reload
 -    Fixes to reload the images of Footer and Background components
       - Images of Footer and Background components doesn't reload with same uri
          so this patch adds global Date() as param in image  url to reload it
    

# Submitter Checklist

review them:

### DIFF
--- a/ui/src/components/Background/index.tsx
+++ b/ui/src/components/Background/index.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { Text, TextContent, TextVariants } from '@patternfly/react-core';
+import background from '../../assets/logo/background.jpg';
 import './Background.css';
 
 const Background: React.FC = () => (
-  <div className="hub-background-image">
+  <div className="hub-background-image" style={{ backgroundImage: `url(${background})` }}>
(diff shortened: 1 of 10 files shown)


ui/src/components/Background/index.tsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
 import { Text, TextContent, TextVariants } from '@patternfly/react-core';
+import background from '../../assets/logo/background.jpg';
```
PuneetPunamiya, line 3: nit: 
```suggestion
import Background from '../../assets/logo/background.jpg';
```

PuneetPunamiya (commented):
Thanks for the pr :call_me_hand: 

Just one thing `package-lock.json`  gets updated could you please check that once

vinamra28:
> Thanks for the pr 
> 
> Just one thing `package-lock.json` gets updated could you please check that once

For me it's not changing

PR #238, opened 2021-04-09
Fixes CI and catalog refresh test

### LINKED ISSUE #190: Hub CI is passing even if CLI tests fails
# Expected Behavior
CI should fail
# Actual Behavior
CI is passing
Screenshot from 2021-02-11 17-01-25

sm43:
# Changes

Fixes catalog refresh test

The order of catalogs in response depend upon how catalog are added
in db and the execution, so checking the catalog name in expected
array instead of directly by name.

Fixes CI

- CI doesn't use to fail when test fails because sub shell was
not able to terminate main script.
Fixes #190

- It used to run build-test, unit-test and intergration-test in
all the CI jobs, now it runs them in respective jobs.

# Submitter Checklist

review them:


test/presubmit-tests.sh
```diff
@@ -116,7 +129,7 @@ ui-build(){
 
 ### presubmit hooks ###
 
-run_build_tests() {
+build_tests() {
```
PuneetPunamiya, line 132: Any reason we are changing the name of the function

test/presubmit-tests.sh
```diff
@@ -126,7 +139,7 @@ run_build_tests() {
   )
 }
 
-run_unit_tests() {
+unit_tests() {
```
PuneetPunamiya, line 142: same as above

test/presubmit-tests.sh
```diff
@@ -54,8 +55,12 @@ install-node() {
 
 ui-unittest(){
     install-node
+    npm install
```
PuneetPunamiya, line 58: Not sure if we need `npm install` before unit test

api/pkg/service/catalog/catalog_test.go
```diff
@@ -116,11 +115,8 @@ func TestRefresh_All(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 2, len(jobs))
 
-	assert.Equal(t, uint(10001), jobs[0].ID)
```
PuneetPunamiya, line 119: Are these tests failing ??

sm43: yes

test/presubmit-tests.sh
sm43: yes it is needed otherwise npm test fails

test/presubmit-tests.sh
sm43: yes build test job in ci needs the func to be `build_tests` other wise it will run all func

test/presubmit-tests.sh
sm43: yes same as above 😆

test/presubmit-tests.sh
vinamra28, line 58: In case of CI we should prefer `npm clean-install` instead of `npm install`

test/presubmit-tests.sh
```diff
@@ -85,20 +90,28 @@ api-unittest(){
...
   info Running Yamllint
 
-  yamllint -c .yamllint ./config.yaml ./config
+  yamllint -c .yamllint ./config.yaml ./config  || {
+    err 'yaml lint failed'
+    return 1
+  }
 }
 
 api-build(){
```
vinamra28, line 117:
```
go: downloading github.com/tektoncd/plumbing v0.0.0-20210408102848-e6e08917cd14
pattern ./cmd/...: lstat ./cmd/: no such file or directory
```
I guess we need to add `cd API_DIR`

vinamra28 (changes requested):

piyush-garg:
Just what @vinamra28 commented, else lgtm

vinamra28 (commented):
I am not sure whether it's a good practice to remove the files from `vendor/github.com/tektoncd/plumbing` and only keep the files present in `vendor/github.com/tektoncd/plumbing/scripts` since we don't need them and earlier also we were not having those files. Rest the PR looks good
/lgtm
/cc @piyush-garg

sm43:
> I am not sure whether it's a good practice to remove the files from `vendor/github.com/tektoncd/plumbing` and only keep the files present in `vendor/github.com/tektoncd/plumbing/scripts` since we don't need them and earlier also we were not having those files. Rest the PR looks good
> /lgtm
> /cc @piyush-garg

PuneetPunamiya:
I think we'll have to add a function for ui lint check as well because in the docs it says it is checking the lint but locally it is not showing any 
For ref : https://create-react-app.dev/docs/running-tests/#on-your-own-environment last point

piyush-garg:
> I am not sure whether it's a good practice to remove the files from `vendor/github.com/tektoncd/plumbing` and only keep the files present in `vendor/github.com/tektoncd/plumbing/scripts` since we don't need them and earlier also we were not having those files. Rest the PR looks good
> /lgtm
> /cc @piyush-garg

We should not do any manual changes with the vendor. If we do it, then you will see the changes every time you will do `go mod vendor` on your local system/ dev flow etc

Issue #252, opened 2021-04-23
Typing on search is reflected only after a short delay


sthaha:
# Expected Behavior
The text you typed should show up immediately on the search input 

# Actual Behavior

The text you typed shows up after a delay (~300ms)

![Kapture 2021-04-23 at 16 37 24](https://user-images.githubusercontent.com/3005132/115831584-21bad580-a455-11eb-889a-5357c979d9a0.gif)

# Steps to Reproduce the Problem

1. Click on search
2. type something
3. The text is reflected only after ~200ms

PR #259, opened 2021-05-11
UI: Fixes search bar and debounce for search

pratap0007:
This patch fixes following
    
   - Previously search bar was showing up searched text after
      a delay(~300 ms),now search bar gets reflected immediate
      while searching a resource
    
  - Previously debounce was not working as expected and now it is fixed
      with help of `use-debounce` npm package and removes useDebounce component
      from utils directory, now users will  get filtered resources just
      after 500 ms when they stop typing in the search bar
    
Closes: #252 

# Submitter Checklist

review them:


ui/src/containers/Search/index.tsx
```diff
@@ -18,11 +25,22 @@ const Search: React.FC = () => {
...
-  const inputRef = React.useRef<HTMLInputElement>(null);
+  const onSearchChange = () => {
+    const currentInput = inputRef.current;
+    assert(currentInput);
+    setValue(currentInput.value);
+    debounced(currentInput.value);
+  };
 
-  const onSearchChange = useDebounce(resources.search, 400);
+  const inputRef = React.useRef<HTMLInputElement>(null);
```
PuneetPunamiya, line 43: Can you move this above the `onSearchChange` function

ui/src/containers/Search/index.tsx
```diff
@@ -18,11 +25,22 @@ const Search: React.FC = () => {
     }, 20);
   });
 
-  const { resources } = useMst();
+  const debounced = useDebouncedCallback(
```
PuneetPunamiya, line 28: Could you add a comment here what this function does ??

ui/src/containers/Search/index.tsx
```diff
@@ -18,11 +25,22 @@ const Search: React.FC = () => {
...
+    (value: string) => {
+      resources.setSearch(value);
+    },
+    // delay in ms
```
PuneetPunamiya, line 32: nit: 
```suggestion
    // Delay in ms
```

ui/src/containers/Search/index.tsx
```diff
@@ -18,11 +25,22 @@ const Search: React.FC = () => {
...
+    500
+  );
 
-  const inputRef = React.useRef<HTMLInputElement>(null);
+  const onSearchChange = () => {
+    const currentInput = inputRef.current;
+    assert(currentInput);
```
PuneetPunamiya, line 38: Can we just assert like this 
```suggestion
    assert(inputRed.current);
```

PuneetPunamiya (commented):
Thanks for the patch!
Could you please address the minor comments

ui/src/containers/Search/index.tsx
pratap0007: yes sure

ui/src/containers/App/__snapshots__/App.test.tsx.snap
```diff
@@ -29,13 +29,13 @@ exports[`App should render the component correctly and match the snapshot 1`] =
                           <div className=\\"pf-l-grid\\">
                             <GridItem span={11}>
                               <div className=\\"pf-l-grid__item pf-m-11-col\\">
-                                <Search>
-                                  <ForwardRef value=\\"\\" type=\\"search\\" onChange={[Function: onChange]} onKeyPress={[Function: onSearchKeyPress]} aria-label=\\"text input example\\" placeholder=\\"Search for resources...\\" spellCheck=\\"false\\" className=\\"hub-search\\">
-                                    <TextInputBase value=\\"\\" type=\\"search\\" onChange={[Function: onChange]} onKeyPress={[Function: onSearchKeyPress]} aria-label=\\"text input example\\" placeholder=\\"Search for resources...\\" spellCheck=\\"false\\" className=\\"hub-search\\" innerRef={{...}} isRequired={false} validated=\\"default\\" isDisabled={false} isReadOnly={false} isLeftTruncated={false}>
+                                <Memo(wrappedComponent)>
```
PuneetPunamiya, line 32: Is this change in test expected here ??

ui/src/containers/Search/index.tsx
```diff
@@ -1,28 +1,48 @@
...
+  /* Debounce callback
+     This function delay the invoking func for 500ms after
+     when users stop searching
+  */
+  const debounced = useDebouncedCallback(
+    (value: string) => {
+      resources.setSearch(value);
+    },
+    // Delay in ms
+    500
```
PuneetPunamiya, line 37: Why the delay in search is 500 ??
Where the user data which he has typed for searching will be stored till 500ms are done? Should that be stored in the localstorage ??

ui/src/containers/App/__snapshots__/App.test.tsx.snap
pratap0007: yes

ui/src/containers/Search/index.tsx
pratap0007: it's a custom delay , I kept it 500 ms (I thought 500 ms is fine ,if you think any other time would be fine we can change it).
search data is fetching from the `search box` itself

PuneetPunamiya: Does it depend on the number of resources in the store ??

pratap0007: nope , it's a delay to used for debounce (any function  will be execute only after that delay), In our case store get update (filter function to be called) only just after 500 ms when user stop searching

vinamra28 (commented):
/lgtm

PR #273, opened 2021-06-07
Updates the Response of Resources API's by adding array of Categories 

PuneetPunamiya:
- This patch updates the design of Resources by adding
  an array of Categories for the following API's

     - /resource/{catalog}/{kind}/{name}/{version}
     - /resource/version/{versionId}
     - /v1/resource/{catalog}/{kind}/{name}/{version}
     - /v1/resource/version/{versionId}

- Updates the resources tests which adds category array in the
  response of CatalogKindNameVersion and ResourceBYVersionID

# Changes

# Submitter Checklist

review them:


api/pkg/shared/resource/resource.go
```diff
@@ -215,6 +215,7 @@ func withResourceVersionDetails(db *gorm.DB) *gorm.DB {
 	return db.
 		Preload("Resource").
 		Preload("Resource.Catalog").
+		Preload("Resource.Category").
```
pratap0007, line 218: WDYT: Ordering of resource's categories as well like resource's tags

sm43: +1 Can we add this?

PuneetPunamiya: Will add it

PR #274, opened 2021-06-09
Update the Catalog Errors order

PuneetPunamiya:
- This patch updates the catalog errors query
    to fetch the errors from db in the specific
    order

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/service/catalog/catalog.go
+++ b/api/pkg/service/catalog/catalog.go
@@ -22,6 +22,7 @@ import (
 	"github.com/tektoncd/hub/api/gen/catalog"
 	"github.com/tektoncd/hub/api/pkg/app"
 	"github.com/tektoncd/hub/api/pkg/db/model"
+	"github.com/tektoncd/hub/api/pkg/parser"
 	"github.com/tektoncd/hub/api/pkg/service/auth"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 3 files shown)


api/pkg/service/catalog/catalog.go
```diff
@@ -139,11 +140,18 @@ func (s *service) CatalogError(ctx context.Context, p *catalog.CatalogErrorPaylo
 		}
 	}
 
+	var errorTypes []string
+
+	for key := range rs {
+		errorTypes = append(errorTypes, key)
+	}
+	sort.Strings(errorTypes)
```
sm43, line 148: error types won't be changing with each call so we can defines file level array
```
var errorTypes = []String{
	parser.Info.String(), ...
}
```
and loop on this

api/pkg/service/catalog/catalog.go
```diff
@@ -139,11 +140,18 @@ func (s *service) CatalogError(ctx context.Context, p *catalog.CatalogErrorPaylo
 		}
 	}
 
+	var errorTypes []string
```
sm43, line 143: also we can remove the switch case, it seems we won't be having error type other than the three as we explicitly add them
```
rs[item.Type] = append(rs[item.Type], item.Detail)
```
this would work

PuneetPunamiya: Yep this should work

api/pkg/service/catalog/catalog.go
```diff
@@ -123,27 +130,22 @@ func (s *service) CatalogError(ctx context.Context, p *catalog.CatalogErrorPaylo
 	}
 
 	allCtgError := []model.CatalogError{}
-	if err := db.Where(&model.CatalogError{CatalogID: ctg.ID}).Find(&allCtgError).Error; err != nil {
+	if err := db.Where(&model.CatalogError{CatalogID: ctg.ID}).Order("id").Find(&allCtgError).Error; err != nil {
```
pratap0007, line 133: I think we can skip Order now , wdyt ?

PR #282, opened 2021-06-29
Add platform information support for resources

barthy1:
# Changes

Following [TEP 70](https://github.com/tektoncd/community/blob/main/teps/0070-tekton-catalog-task-platform-support.md), this code allows the hub to parse, store and return platform information for the resources.

- add platform type to return platform information for resources and resource versions in the APIs
 ```
  - /v1/resources
  - /v1/resource/<id>
  - /v1/resource/<catalog>/<kind>/<name>
  - /v1/resource/<catalog>/<kind>/<name>/<version>
  - /v1/resource/<id>/versions
  - /v1/resource/version/<versionID>
  - /resources
  - /resource/<id>
  - /resource/<catalog>/<kind>/<name>
  - /resource/<catalog>/<kind>/<name>/<version>
  - /resource/<id>/versions
  - /resource/version/<versionID>
```
- update parser to get platform information from catalog resources using "tekton.dev/platforms" annotation
  (if there is no "tekton.dev/platforms" annotation in the resource content, "linux/amd64" value is assigned)
- add tables for platform and many2many relations to resource and resource version

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

# Submitter Checklist

review them:

### CODE REVIEW

File: api/pkg/service/catalog/syncer.go [hunk 1]
```diff
@@ -282,6 +283,28 @@ func (s *syncer) updateResourceTags(
...
+func (s *syncer) updateResourcePlatforms(
+	txn *gorm.DB, log *zap.SugaredLogger,
+	res *model.Resource, platforms []string) {
+
+	if len(platforms) == 0 {
+		return
+	}
+
+	// clean existing platforms
+	txn.Unscoped().Where(&model.ResourcePlatform{ResourceID: res.ID}).Delete(&model.ResourcePlatform{})
```
sm43, line 295: do we need to clean the resources before adding?

sm43: I think `FirstOrCreate` will create record only if it is missing, so what if we don't clean then?

api/test/fixtures/platforms.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright © 2020 The Tekton Authors.
```
sm43, line 1: nit: 2021

api/pkg/shared/resource/resource.go
```diff
@@ -47,7 +48,7 @@ var (
 
 // Query resources based on name, kind, tags.
 // Match is the type of search: 'exact' or 'contains'
-// Fields: name, []kinds,[]Catalogs []Tags, Match, Limit
+// Fields: name, []kinds,[]Catalogs []Tags, []Platforms, Match, Limit
```
sm43, line 51: could you add a test for querying using `platforms`?

api/test/fixtures/resource_platforms.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright © 2020 The Tekton Authors.
```
PuneetPunamiya, line 1: nit: 2021

api/test/fixtures/platforms.yaml
barthy1: fixed

api/pkg/shared/resource/resource.go
barthy1: done

api/pkg/service/catalog/syncer.go
barthy1: In case if we don't clean platforms, new platforms will be just added to the existing ones (when it's necessary). This doesn't cover the use case when some platform was removed from the list of supported platforms in the resource.
Clean up of the existing records and addition of the new list would represent the actual data from the resource.

api/test/fixtures/resource_platforms.yaml
barthy1: fixed

sm43:
@barthy1 pr looks good to me. a few ques though 😅 
we are adding platform as a property of [resource](https://github.com/tektoncd/hub/blob/fb4fdf48c9225ae7dcc871a58d4d755daf10ca8f/api/pkg/db/model/model.go#L73)
wouldn't this imply that all versions of the resources work on those platform?

 Lets say `v0.1` of a resource `r` is supported on `x`, `y` platform
 then would `v0.2` and further versions will also be supported/works on both?
 If yes then keeping it as resource property would be fine.
 but if not always then
how would a user knows whether version `v0.2`, `v0.3`, ... are supported or not?

I think `platform` could be added as a property of resource's version.
because if the resource object says it is supported on `x`, `y` and a future version of resource is not supported on `y`
then it would be a bad experience for user.

- I think we should add `platform` in resource version  
- but I also this feel the list of platform should in resource object which would say some or all version of this resource works on this platforms

WDYT?

barthy1:
@sm43 Thank you for going deeper into this topic!
Yes, you are correct, different resource versions can work on different plaforms. However I used as an example approach which hub has for tags, my understanding that different versions can have different tags, but still tags are listed on resource level.
Thinking about it more, I believe that in case of platforms ability to stick platforms to version is more important than for tags...
So, yes, it makes sense to have platforms on resource version level, I will work on it after my vacation.

Regarding platforms on resource level, I agree, that resource object should provide some info about platforms, as for some API requests, resource version objects are not included.. I was thinking to use latestVersion, however it is also used not everywhere. So yes, probably it makes sense to have separate platforms field in resource object, generated from platforms of all versions.

sm43:
>   approach which hub has for tags, my understanding that different versions can have different tags, but still tags are listed on resource level.

yep. this does came to my mind but tags are kind of metadata so I think it wouldn't affect user much  

>  I believe that in case of platforms ability to stick platforms to version is more important than for tags..
 So, yes, it makes sense to have platforms on resource version level, I will work on it after my vacation.

yep 😃  👍🏽 
> So, yes, it makes sense to have platforms on resource version level, I will work on it after my vacation.
> Regarding platforms on resource level, I agree, that resource object should provide some info about platforms, as for some API requests, resource version objects are not included.. I was thinking to use latestVersion, however it is also used not everywhere. So yes, probably it makes sense to have separate platforms field in resource object, generated from platforms of all versions.

yeah we can think about how we can include platforms in resource obj

barthy1:
@sm43 could please have a look one more time? I've added Platforms on resource and resource version levels. Data is gathered and stored during catalog sync. Platforms in Resource object is just list with unique platforms from resource versions.

api/pkg/db/model/model.go
```diff
@@ -33,6 +33,13 @@ type (
 		Resources []*Resource `gorm:"many2many:resource_tags;"`
 	}
 
+	Platform struct {
+		gorm.Model
+		Name             string             `gorm:"not null;unique"`
+		ResourceVersions []*ResourceVersion `gorm:"many2many:resource_platforms;"`
+		Resource         []*Resource        `gorm:"many2many:version_platforms;"`
```
sm43, line 40: mismatch I guess :)
`version_platforms` for `Resource` instead of `resource_platform`

api/pkg/service/catalog/syncer.go
```diff
@@ -262,8 +262,10 @@ func (s *syncer) updateResources(
 
 		rerr.Combine(s.updateResourceCategory(txn, log, &dbRes, r.Categories))
 		s.updateResourceTags(txn, log, &dbRes, r.Tags)
-		s.updateResourceVersions(txn, log, catalog, dbRes.ID, r.Versions)
-
+		// platform ids on resource version level
+		verPlatformIds := map[uint]bool{}
+		s.updateResourceVersions(txn, log, catalog, dbRes.ID, r.Versions, &verPlatformIds)
```
sm43, line 267: if we add platforms only for resource versions then I think we can fetch platforms for a resource from its versions.
i think in gorm queries is we can use https://gorm.io/docs/preload.html#Nested-Preloading and then get platforms for a resource

sm43: but this also means we will have to loop on all resource versions everytime we want to fetch resource details. I am okay with adding for both but wdyt ? which would be better?

api/pkg/service/catalog/syncer.go
```diff
@@ -288,6 +290,24 @@ func (s *syncer) updateResourceTags(
...
+	for verPlatformId := range verPlatformIds {
+		platformIds = append(platformIds, verPlatformId)
+		platform := model.Platform{}
+		txn.First(&platform, verPlatformId)
+		resPlatform := model.ResourcePlatform{ResourceID: res.ID, PlatformID: verPlatformId}
+		txn.Model(&model.ResourcePlatform{}).Where(resPlatform).FirstOrCreate(&resPlatform)
+		log.Infof("Resource: %d: %s | platform: %s (%d)", res.ID, res.Name, platform.Name, platform.ID)
+	}
+	// remove mapping for the platforms, which are not in the list
+	txn.Unscoped().Where(&model.ResourcePlatform{ResourceID: res.ID}).
```
sm43, line 306: +1

sm43:
@barthy1 Thank you. this looks nice just have a couple of comments :)

api/pkg/service/catalog/syncer.go
barthy1: I tried several approaches here. 
The first idea was to have 1 join table for resources, versions and platforms, however at the end the problem was to get unique platforms for resources, because by default I got the same values several times in case of simple Preload, and Distinct was not possible to add to the default Preload, and at the end I thought that it's too complicated modifications in too many places.
The second idea was similar to your suggestion with getting resource platforms from resource versions' platforms, and at the end I decided not to implement it, because some queries work on resource level only, others are designed to return result of one db.Select and we would really need to extend many requests and loop through all resource versions all the time. So I thought that's too much overhead here.
That's why I chose the third way - separate storage of platforms for resources and platforms for resource versions. As we need to go through resource versions only when there are changes on catalog side and write or delete is done only when it's necessary. + the new logic is isolated.

However here I'd rely on your experience and project roadmap understanding. What is your gut feeling? :)

api/pkg/db/model/model.go
barthy1: Thank you and good catch :)

api/pkg/db/model/model.go
```diff
@@ -63,7 +70,8 @@ type (
 		Categories []*Category `gorm:"many2many:resource_categories;"`
 		CatalogID  uint
 		Versions   []ResourceVersion
-		Tags       []*Tag `gorm:"many2many:resource_tags;"`
+		Tags       []*Tag      `gorm:"many2many:resource_tags;"`
+		Platforms  []*Platform `gorm:"many2many:resource_platforms;"`
```
pratap0007, line 74: Could you please add `OnDelete:CASCADE` constraints as well

api/pkg/db/model/model.go
```diff
@@ -75,6 +83,7 @@ type (
 		MinPipelinesVersion string `gorm:"not null;default:null"`
 		Resource            Resource
 		ResourceID          uint
+		Platforms           []*Platform `gorm:"many2many:version_platforms;"`
```
pratap0007, line 86: and same here

api/pkg/service/catalog/syncer.go
sm43: oh yeah in first case there would be redundancy 
>db.Select and we would really need to extend many requests and loop through all resource versions all the time. So I thought that's too much overhead here.

yep that's was my concern

>However here I'd rely on your experience and project roadmap understanding. What is your gut feeling? :)

I would like to keep the current that's seems better, we can improve it later if found a better way :)

api/pkg/db/model/model.go
barthy1: @pratap0007 constraints we added. Could you please add `lgtm` if it looks ok to you?

pratap0007:
PR looks good to me cc @PuneetPunamiya @piyush-garg

PR #284, opened 2021-07-02
Parses deprecation info from resources and update the API's Response

pratap0007:
# Changes

This patch includes  

 * Parse deprecation info of resources and fill into db    
    - Parse deprecation info from versions of resources
    - Add migration to add a column to store deprecation info
 * Updates the Response of API's to include deprecation info for a version 
    - If a version/latest version of resource is deprecated then it adds
      deprecation info to respective version info in the response as
      `deprecated:"true"` otherwise it doesn't  add
    
    - Following api's response would change
    
            - /resource/version/<versionID>
            - /resource/version/<version>
            - /resource or /query (if latest version is deprecated)
            - v1/resource/version/<versionID>
            - v1/resource/version/<version>
            - v1/resources or /query (if latest version is deprecated)
* Updates golden file for search command in cli

# Submitter Checklist

review them:


api/pkg/parser/catalog.go
```diff
@@ -264,6 +266,14 @@ func (c CatalogParser) appendVersion(res *Resource, filePath string) Result {
 		result.Info(issue)
 	}
 
+	// optional check
+	deprecated, ok := annotations[DeprecatedAnnotation]
+	if ok {
+		issue := fmt.Sprintf("Resource %s - %s version of %s has been deprecated", tkn.GVK, version, tkn.Name)
+		log.With("action", "ignore").Info(issue)
+		result.Info(issue)
```
sm43, line 274: wouldn't this add to `catalogerror` table?

sm43: do we need to add that record?

sm43:
could you add few tests for the deprecated case?

api/pkg/parser/catalog.go
pratap0007: not sure on that ,have added just for info

api/pkg/service/resource/resource.go
```diff
@@ -297,6 +297,12 @@ func initResource(r model.Resource) *resource.ResourceData {
 		RawURL:              getStringReplacer(lv.URL).Replace(lv.URL),
 		UpdatedAt:           lv.ModifiedAt.UTC().Format(time.RFC3339),
 	}
+
+	// If latest version is deprecated then add it only
+	if lv.Deprecated == "true" {
```
piyush-garg, line 302: Does this mean like if the resource is deprecated then this field will be there in response?

pratap0007: yeah

api/pkg/db/model/model.go
```diff
@@ -80,6 +80,7 @@ type (
 		Description         string
 		URL                 string `gorm:"not null;default:null"`
 		DisplayName         string
+		Deprecated          string
```
sm43, line 83: is it possible to use `bool` type here?

api/design/types/type.go
```diff
@@ -102,6 +102,9 @@ var ResourceVersionData = ResultType("application/vnd.hub.resource.version.data"
 	Attribute("displayName", String, "Display name of version", func() {
 		Example("displayName", "Buildah")
 	})
+	Attribute("deprecated", String, "Deprecation status of a version", func() {
```
sm43, line 105: `Boolean` type here? for `deprecated`?

api/pkg/db/model/model.go
pratap0007: yeah  it's possible,used string as while parsing we get it as string("true") from resources

api/design/types/type.go
pratap0007: it's a string

api/pkg/db/model/model.go
sm43: yeah but boolean in db will take less space and we won't any value other than T/F for now

sm43 (approved):
Thank you for the changes :)

api/pkg/cli/cmd/search/testdata/TestSearch_JSONFormat.golden
```diff
@@ -15,6 +15,7 @@
 				"ID": 12,
 				"Version": "0.2",
 				"DisplayName": "foo-bar-0.1",
+				"Deprecated": null,
```
PuneetPunamiya, line 18: I think the default value should be `false`  ?
cc @pratap0007 @sm43

api/pkg/db/model/model.go
```diff
@@ -80,6 +80,7 @@ type (
 		Description         string
 		URL                 string `gorm:"not null;default:null"`
 		DisplayName         string
+		Deprecated          bool   `gorm:"default:false"`
```
PuneetPunamiya, line 83: In the table we are setting it to false right ? So above instead of `null` should we return `false`

api/pkg/service/resource/resource.go
```diff
@@ -305,6 +305,12 @@ func initResource(r model.Resource) *resource.ResourceData {
 		UpdatedAt:           lv.ModifiedAt.UTC().Format(time.RFC3339),
 		Platforms:           platforms,
 	}
+
+	// If latest version is deprecated then add it only
```
PuneetPunamiya, line 309: Can we please improve the comment message

api/pkg/service/resource/resource.go
```diff
@@ -421,6 +427,11 @@ func versionInfoFromResource(r model.Resource) *resource.ResourceVersion {
 		Platforms:           verPlatforms,
 	}
 
+	// If version is deprecated then add it only
```
PuneetPunamiya, line 430: same as above

api/pkg/service/resource/resource_http_test.go
```diff
@@ -641,3 +641,37 @@ func TestByID_Http_ErrorCase(t *testing.T) {
 		assert.Equal(t, "not-found", err.Name)
 	})
 }
+
+func TestDeprecationByVersionID_Http(t *testing.T) {
```
PuneetPunamiya, line 645: Should we add test when the resource is not deprecated ??

api/v1/service/resource/resource_http_test.go
```diff
@@ -641,3 +641,37 @@ func TestByID_Http_ErrorCase(t *testing.T) {
 		assert.Equal(t, "not-found", err.Name)
 	})
 }
+
+func TestDeprecationByVersionID_Http(t *testing.T) {
```
PuneetPunamiya, line 645: Same as above

PuneetPunamiya (commented):
Thanks for the patch, some nit picks else the pr looks good 👍🏻

api/pkg/cli/cmd/search/testdata/TestSearch_JSONFormat.golden
pratap0007: for Non-deprecated resources deprecated field shouldn't include in the response

api/pkg/service/resource/resource_http_test.go
pratap0007: It's not necessary and that case would be cover in `TestByVersionID_Http` test

api/pkg/cli/cmd/search/testdata/TestSearch_JSONFormat.golden
pratap0007: I think,unspecified field gets nil pointer while unmarshaling hence there is nil

PR #290, opened 2021-07-13
Adds a migration to remove existing Categories

PuneetPunamiya:
Initially there were some categories in the config.yaml which are not 
part of the predefined category list and thus the existing categories 
were still present in the db, for example there was one category
named `Deploy` which in predefined category list is changed to `Deployment`, 
hence in order to keep the list generalized and have proper resource category
mapping and show only the predefined generalized categories on UI, this patch 
adds a migration to remove this existing categories 

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/db/migration/202107131802_remove_existing_categories.go
+++ b/api/pkg/db/migration/202107131802_remove_existing_categories.go
@@ -1,0 +1,38 @@
+// Copyright © 2021 The Tekton Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migration
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/tektoncd/hub/api/gen/log"
+	"github.com/tektoncd/hub/api/pkg/db/model"
+	"gorm.io/gorm"
+)
+
+// Removes Existing Categories which are not present
+// in pre-defined Category List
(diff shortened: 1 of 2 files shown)


api/pkg/db/migration/202107131802_remove_existing_categories.go
```diff
@@ -0,0 +1,44 @@
...
+	return &gormigrate.Migration{
+		ID: "202107131802_remove_existing_categories",
+		Migrate: func(db *gorm.DB) error {
+			var cat []model.Category
+			if err := db.Where("name IN ?", []string{"Deploy", "Editor", "Language", "Notification", "Others", "Test Framework"}).Find(&cat).Error; err != nil {
+				log.Error(err)
+				return err
+			}
+
+			if err := db.Delete(cat).Error; err != nil {
```
sm43, line 37:
```
db.Where("name IN ?", []string{"Deploy", "Editor", "Language", "Notification", "Others", "Test Framework"}).Delete(&User{})
```
would this work?

PuneetPunamiya: yep this would work as well, I thought of getting the records first and then updating the deleted at, but this approach looks fine

api/pkg/db/migration/202107131802_remove_existing_categories.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Removes Existing Categories which are not present
+// in pre-defined Category List
+func removeExistingCategories(log *log.Logger) *gormigrate.Migration {
+
+	return &gormigrate.Migration{
+		ID: "202107131802_remove_existing_categories",
+		Migrate: func(db *gorm.DB) error {
+			if err := db.Where("name IN ?", []string{"Deploy", "Editor", "Language", "Notification", "Others", "Test Framework"}).Delete(&model.Category{}).Error; err != nil {
```
pratap0007, line 31: could you please delete it permanently

PuneetPunamiya: Gorm does a soft delete and with soft delete the entry will be just there in the db with the `delete_at` time added for that particular record, so in any of the response we won't get these entries from the db

pratap0007: but entries would be remain in the table, Gorm does permanent delete as well

PR #292, opened 2021-07-15
Update the Catalog sync to delete the removed resources info from existing db

pratap0007:
This patch includes following changes

- Adds `OnDelete:CASECADE` constraint on resource_tags and
  resource_categories table
- Update the catalog sync which compares hosted catalog resources to
  existing db info of resources for same catalog while syncing catalogs and delete
  resources info from the db if any resources from the hosted catalog are removed

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/service/catalog/syncer.go
+++ b/api/pkg/service/catalog/syncer.go
@@ -208,6 +208,11 @@ func (s *syncer) updateJob(syncJob model.SyncJob, sha string, res []parser.Resou
 	}
 	catalog.SHA = sha
 
+	if err := s.deleteDBResources(txn, log, &catalog, res); err != nil {
+		txn.Rollback()
+		return err
+	}
+
 	rerr, err := s.updateResources(txn, log, &catalog, res)
 	if err != nil {
 		txn.Rollback()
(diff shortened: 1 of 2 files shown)


api/pkg/service/catalog/syncer.go
```diff
@@ -404,3 +409,56 @@ func (s *syncer) updateCatalogResult(
 	}
 	return nil
 }
+
+// Delete the resource by id
+func (s *syncer) deleteResource(txn *gorm.DB, log *zap.SugaredLogger, res model.Resource) {
+
+	txn.Unscoped().Where("id = ?", res.ID).Delete(&model.Resource{})
```
sm43, line 416: on what condition are we deleting resource?

sm43: deleting all and adding again?

pratap0007: it compares db resources and sync resources by `Kind and Name` and find all resources which are in db and not in sync resources list and then deletes one by one resource

sm43: ah okay

sm43: what if we add all resources in db as current flow and save the id of resources in array and then use that to delete resources?
delete resources whose id is not in the array

sm43: similar to your approach
wdyt

pratap0007: yes,it would be a better approach than current

api/pkg/db/migration/migration.go
```diff
@@ -55,9 +56,6 @@ func Migrate(api *app.APIBase) error {
 			&model.SyncJob{},
 			&model.Scope{},
 			&model.Config{},
-			&model.Platform{},
```
piyush-garg, line 58: Why are we removing these from here ?

pratap0007: These were added  in the migration as well

sm43: why are removing `platform`? the other 2 will be autocreated how platform will get created on a new db installation if we remove from here?

api/pkg/service/catalog/syncer.go
```diff
@@ -405,3 +429,39 @@ func (s *syncer) updateCatalogResult(
...
+
+	log.Infof("Resource %s of kind %s has been deleted", res.Name, res.Kind)
+}
+
+// Delete the tags which doesn't belongs to any other existing resources
+func (s *syncer) deleteTag(txn *gorm.DB, log *zap.SugaredLogger, tags []model.ResourceTag) {
```
sm43, line 442: do we need to do this as part of this api ? 🤔  what if we have a clean up api which would remove things like this which are no longer required

api/pkg/db/migration/migration.go
pratap0007: here platforms is optional and we can keep or remove it,if we remove platforms from here it will get created by the automigration on a new db installation

api/pkg/service/catalog/syncer.go
pratap0007: I think yes , if we want to have a clean up api then how we would identify which things are no longer required ?

api/pkg/db/migration/migration.go
sm43: okay but I didn't understood why this is optional? 🤔 on new db creation `initschema` https://github.com/tektoncd/hub/blob/3c1ec7d356748afce2281879662989d5d8fb3d10/api/pkg/db/migration/migration.go#L47  is run right?
and if platforms is not added there still it will get created

pratap0007: yes

api/pkg/db/migration/202107221500_add_ondelete_constraints.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Drops existing foreign key constraints and create the same constraints
+// in order to add another OnDelete constraints to foriegn key
+func addOnDeleteConstraints(log *log.Logger) *gormigrate.Migration {
+
+	return &gormigrate.Migration{
+		ID: "202107221500_add_ondelete_constraints",
+		Migrate: func(db *gorm.DB) error {
+			txn := db.Begin()
+
```
PuneetPunamiya, line 32: Can you create a function like this https://github.com/tektoncd/hub/blob/main/api/pkg/db/migration/202107291608_refresh_all_tables.go#L36
and handle the error in a different function instead of writing `txn.Rollback` multiple times

api/pkg/db/migration/202107221500_add_ondelete_constraints.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func addOnDelete(txn *gorm.DB, log *log.Logger) error {
+	if err := txn.Migrator().DropConstraint(&model.ResourceCategory{}, "fk_resource_categories_resource"); err != nil {
+		log.Error(err)
+		return err
+	}
+
+	if err := txn.Exec("ALTER TABLE resource_categories ADD CONSTRAINT fk_resource_categories_resource  FOREIGN KEY (resource_id) REFERENCES resources(id) ON DELETE CASCADE;").Error; err != nil {
```
PuneetPunamiya, line 51: Not mandatory just a thought can't we use update query from gorm instead of writing the whole query, else the patch looks fine to me

PuneetPunamiya: For example 
Drop Constraint 
update column from table by adding foreign key constraints with on Delete cascade 
For ref : https://gorm.io/docs/update.html#Update-single-column

pratap0007: No, we can't do with gorm and with gorm CreateConstraint does't create

api/pkg/db/migration/migration.go
PuneetPunamiya: So platform table is getting created because of this migration https://github.com/tektoncd/hub/blob/cc843047e14327e18a8d12de66083cbcb3b5145b/api/pkg/db/migration/202106281500_create_platform_tables.go#L24

pratap0007:
@vinamra28 done, could you please take a look once more

PR #301, opened 2021-07-30
Refeshes all tables in Hub DB

PuneetPunamiya:
- This patch collects the existing data from db deletes
   the existing tables, recreates the tables back and
   inserts the data in order to update the db and makes it
   compatible with `gorm 2.0`

# Changes

# Submitter Checklist

review them:


vinamra28 (approved):
/lgtm

api/pkg/db/migration/migration.go
```diff
@@ -36,9 +36,8 @@ func Migrate(api *app.APIBase) error {
 			addRefreshTokenChecksumColumnInUserTable(log),
 			updateCatalogBranchToMain(log),
 			addAvatarURLColumnInUsersTable(log),
-			removeCatgoryIdColumnAndConstraintsFromTagtable(log),
+			refreshAllTables(log),
 			updateResourcesCategoryTable(log),
-			removeExistingCategories(log),
```
sm43, line 41: why are we removing migrations?

sm43: don't we need to run them after refreshing the db?

sm43: refreshAllTables this migration has to be run before all newly added migration, after that the remaning will be executed properly

vinamra28: nope.... refresh_all_tables migration will do the job for us

api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+
+			var catalog_error []model.CatalogError
+			if err := db.Find(&catalog_error).Error; err != nil {
+				log.Error(err)
+				return err
+			}
+
+			txn := db.Begin()
+			if err := db.Migrator().DropTable(&model.Category{},
```
sm43, line 101: we are starting a `txn` but not using them while executing queries then how it would be helpful 😸

api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -0,0 +1,262 @@
...
+					return err
+				}
+			}
+
+			if len(catalog_error) > 0 {
+				if err := db.Create(
+					&catalog_error,
+				).Error; err != nil {
+					log.Error(err)
+					txn.Rollback()
```
sm43, line 154: we can move all queries in a func and handle the rollback from there instead of adding for each query

api/pkg/db/migration/202107291608_refresh_all_tables.go
sm43: https://gorm.io/docs/transactions.html#Transactions-by-manual

api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -0,0 +1,262 @@
...
+				&model.UserScope{},
+				&model.Config{},
+				&model.ResourceTag{},
+			); err != nil {
+				log.Error(err)
+				txn.Rollback()
+				return err
+			}
+
+			if len(catalog) > 0 {
```
sm43, line 139: why do we need `> 0` check?

sm43: oh may be while setting up a new instance

api/pkg/db/migration/202107291608_refresh_all_tables.go
sm43:
```
txn := db.Begin()
if err:= refreshDb(txn, log); err!=nil{
  log error
  txn.rollback()
}
txn.commit
```
and use txn passed for all queries

api/pkg/db/migration/202107291608_refresh_all_tables.go
vinamra28: just handling the corner case if in case any table is having any zero values

api/pkg/db/migration/migration.go
sm43: did you test this on a cluster?

vinamra28: yes

api/pkg/db/migration/202107131802_remove_existing_categories.go
```diff
@@ -28,7 +28,7 @@ func removeExistingCategories(log *log.Logger) *gormigrate.Migration {
 	return &gormigrate.Migration{
 		ID: "202107131802_remove_existing_categories",
 		Migrate: func(db *gorm.DB) error {
-			if err := db.Unscoped().Where("name IN ?", []string{"Deploy", "Editor", "Language", "Notification", "Others", "Test Framework"}).Delete(&model.Category{}).Error; err != nil {
+			if err := db.Migrator().DropTable(&model.Category{}); err != nil {
```
pratap0007, line 31: I think, we should delete all category instead of dropping the table

api/pkg/db/migration/202107131802_remove_existing_categories.go
```diff
@@ -28,7 +27,7 @@ func removeExistingCategories(log *log.Logger) *gormigrate.Migration {
 	return &gormigrate.Migration{
 		ID: "202107131802_remove_existing_categories",
 		Migrate: func(db *gorm.DB) error {
-			if err := db.Unscoped().Where("name IN ?", []string{"Deploy", "Editor", "Language", "Notification", "Others", "Test Framework"}).Delete(&model.Category{}).Error; err != nil {
+			if err := db.Exec("DELETE FROM categories").Error; err != nil {
```
sm43, line 30: I think we can remove this migration now. as we are updating the config refresh api to handle deletion
??

api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -0,0 +1,274 @@
...
+		&model.Scope{},
+		&model.UserScope{},
+		&model.Config{},
+		&model.ResourceTag{},
+	); err != nil {
+		log.Error(err)
+		return err
+	}
+
+	if len(catalog) > 0 {
```
sm43, line 157: just curious when we fetch records and insert as it is ... do the creation timestamp changes ? or remains same as before

api/test/fixtures/sync_jobs.yaml
```diff
@@ -15,11 +15,12 @@
 # NOTE:
 # This record is added to make sure the sync_jobs table state remains same
 # for every catalog refresh test. not adding this will fails the test due to the
-# sequence which will be out of sync and have to manually bring it back in sync. 
+# sequence which will be out of sync and have to manually bring it back in sync.
 # Error: ERROR: duplicate key value violates unique constraint "sync_jobs_pkey"
 
 - id: 1
```
sm43, line 21: jfyi the test didn't failed due to this record because we don't use this record in test.
see the note above :)

sm43 (changes requested):

api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -0,0 +1,274 @@
...
+package migration
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"github.com/tektoncd/hub/api/gen/log"
+	"github.com/tektoncd/hub/api/pkg/db/model"
+	"gorm.io/gorm"
+)
+
+func refreshAllTables(log *log.Logger) *gormigrate.Migration {
```
sm43, line 24: can you add a desc for this why this is required
and what is this taking care of and the why we dont' require the migration for eg `removeCatgoryIdColumnAndConstraintsFromTagtable` now

api/pkg/db/migration/202107131802_remove_existing_categories.go
PuneetPunamiya: But before running the api we need to fix the categories as well right ??

api/pkg/db/migration/202107291608_refresh_all_tables.go
PuneetPunamiya: I think it should remain the same since we are inserting the data which we have collected from the db itself and also the id for all is not changed so I think it should remain the same

api/pkg/db/migration/202107131802_remove_existing_categories.go
sm43: to avoid deletion of all records which are not necessarily to be deleted we are updating the config refresh api
so refreshall migration will add those back in db
and when api server is started the new will get added the ones removed from config will get deleted

PuneetPunamiya: Yeah agreed on that but ones removed from config which are supposed to be deleted needs to be handled at the api level itself then right

api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -0,0 +1,279 @@
...
+	return &gormigrate.Migration{
+		ID: "202107291608_refresh_all_tables",
+		Migrate: func(db *gorm.DB) error {
+			txn := db.Begin()
+			err := migrateDB(txn, log)
+			if err != nil {
+				txn.Rollback()
+				log.Error(err)
+			}
```
sm43, line 38: `return` statement missing

vinamra28 (commented):
/lgtm

PR #303, opened 2021-08-04
API: Updates `/system/config/refresh/` api to delete the removed categories

pratap0007:
- If existing categories are removed from the config file then this patch
  deletes categories and their mapping with the resources from the database
- Deletes migration of removing existing categories as deletion of categories
  would be handled by the config refresh api
Sigend-off-by: Shiv Verma <shverma@redhat.com>

# Changes

# Submitter Checklist

review them:


api/pkg/db/initializer/initializer.go
```diff
@@ -80,6 +81,40 @@ func (i *Initializer) Run(ctx context.Context) (*model.Config, error) {
 
 type initFn func(*gorm.DB, *log.Logger, *app.Data) error
 
+// Deletes categories and it's mapping with resource from the database
+// which are removed from the config
+func deleteCategories(db *gorm.DB, log *log.Logger, data *app.Data) error {
```
sm43, line 86: Instead of doing this separately can we
- after adding the category https://github.com/tektoncd/hub/blob/49e68f837eb3a9f36b9fdbc675bb4d5167c090c1/api/pkg/db/initializer/initializer.go#L87
after this take the id of the category and put in an array
-  so this will add id of all categories which are in the config yaml
- after this delete categories whose id is not in the above array so it will done in a single query

PuneetPunamiya: +1

pratap0007: but we need  to delete resource & category mapping before deleting categories

sm43: yeah can we not use the same resource ids to delete the mapping
delete mapping if resource id not in the array?

pratap0007: yeah we can, but number of query will be increase

sm43:
```
// Not In
db.Not(map[string]interface{}{"name": []string{"jinzhu", "jinzhu 2"}}).Delete(&User{})
// Delete * FROM users WHERE name NOT IN ("jinzhu", "jinzhu 2");
```
wouldn't something like this work?

api/pkg/db/initializer/initializer.go
```diff
@@ -82,13 +82,29 @@ type initFn func(*gorm.DB, *log.Logger, *app.Data) error
...
 	}
+
+	if len(configCatID) > 0 {
+		// Deletes mapping of removed categories and resources from the database
+		if err := db.Unscoped().Not(map[string]interface{}{"category_id": configCatID}).Delete(&model.ResourceCategory{}).Error; err != nil {
+			log.Error(err)
+			return err
+		}
+		// Deletes categories from database which are removed from the config
+		if err := db.Unscoped().Not(map[string]interface{}{"id": configCatID}).Delete(&model.Category{}).Error; err != nil {
```
piyush-garg, line 102: we need to handle the case if the categories is still mapped to any resource

pratap0007: so in that case categories and their mapping with resource shouldn't delete, right ?

sm43: well in that case the the change shouldn't have made in config.yaml
because we are commiting the change in repo and then saying not to delete it from db

sm43: for hub source of truth is config.yaml .. if the category is removed/renamed from config that means we don't want to show that on hub so mapping can be deleted..
when the category is change in catalog.. new mapping with new categories will get added in db

piyush-garg: hmm, cool. So those esources will turn in other categories then right

PuneetPunamiya (commented):
/approve
/meow

PR #304, opened 2021-08-09
Handles the last value of the table while refreshing all tables

PuneetPunamiya:
- While refreshing all the tables in `refresh_all_tables` migration
   it was setting the last value back to 1 where it could create an
   error of duplicate key if a new entry was addded so this patch
   stores the last value of the sequence and while reinserting the
   data it alters the sequence by last_value+1 in order to avoid the
   duplicate key constraint error

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/db/migration/202107291608_refresh_all_tables.go
+++ b/api/pkg/db/migration/202107291608_refresh_all_tables.go
@@ -15,6 +15,8 @@
 package migration
 
 import (
+	"fmt"
+
 	"github.com/go-gormigrate/gormigrate/v2"
 	"github.com/tektoncd/hub/api/gen/log"
 	"github.com/tektoncd/hub/api/pkg/db/model"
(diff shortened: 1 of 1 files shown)


api/pkg/db/migration/202107291608_refresh_all_tables.go
```diff
@@ -167,6 +235,12 @@ func migrateDB(txn *gorm.DB, log *log.Logger) error {
 			log.Error(err)
 			return err
 		}
+
+		query := fmt.Sprintf("ALTER SEQUENCE catalogs_id_seq RESTART WITH %d", catalogs_last_value+1)
```
sm43, line 239: do we need to `+ 1` if we getting previous last value and restart with it back?

PuneetPunamiya: Yes we have to because if there is a new entry then it considers the last value and inserts the value at the last value hence doing +1

sm43: ah okay 👍🏽

pratap0007:
lgtm

piyush-garg:
lgtm cc @vinamra28

vinamra28:
/hold
for sync_jobs this is remaining

PuneetPunamiya:
@vinamra28  Can you take a final look at this once ?

vinamra28:
Tested on cluster as well :+1: 
thanks for fixing this
/lgtm
/meow

Issue #288, opened 2021-07-08
Hub UI displays username in lower case even if actual GitHub username contains both lower and upper case


vinamra28:
# Expected Behavior

Hub UI should display the username in the same way the user has kept in Github. Example my github username is fooBar but Hub UI displays my username as foobar.

# Actual Behavior

Hub UI should display the username in same way we have in Github. example my Github username is fooBar so Hub UI should also display fooBar

# Steps to Reproduce the Problem

1. Have a github username such that it has both lowercase and uppercase letters.
2. Login into Hub UI and see the username on top right corner.

# Additional Info

Probable place where we need to fix is 
https://github.com/tektoncd/hub/blob/9f058519d26936d9eaf8d729cce256e1bae87192/api/pkg/service/auth/auth.go#L135
and the query needs to be updated as
https://github.com/tektoncd/hub/blob/9f058519d26936d9eaf8d729cce256e1bae87192/api/pkg/service/auth/auth.go#L126

/good-first-issue

kk-r:
@vinamra28 Submitted PR - #295  Thanks for all the details. It really helpful!

 I'd like to mention that I have found additional case when username in database is mixed-case and user changed to lowercase in GitHub(as per GitHub `fooBar` and  `foobar` both are same.). So w need to handle this case by lowercase the value in our database and check with username from GitHub. 
 
`Where("LOWER(github_login) = ? or github_login = ?", strings.ToLower(ghUser.GetLogin()), ghUser.GetLogin())`

I'm new to this community. If there is anything missing or need to cover, please let me know. Thanks again! :)

PR #305, opened 2021-08-10
Remove username formatting(convert lowercase) while saving the user

kk-r:
Resolves: #288

# Changes
- Update user exists check logic for both lowercase & mixed-case username
    - Using `LOWER(github_login)` to prevent if user changed his username from mixed-case to lowercase in GitHub.
- Store updated username if the username in database is incorrect.
- Update existing testcase to support when new username contains mixed-case letters.

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/service/auth/auth.go
+++ b/api/pkg/service/auth/auth.go
@@ -132,7 +132,7 @@ func (r *request) addUser(ghUser *github.User) (*model.User, error) {
 		if err == gorm.ErrRecordNotFound {
 
 			user.GithubName = ghUser.GetName()
-			user.GithubLogin = strings.ToLower(ghUser.GetLogin())
+			user.GithubLogin = ghUser.GetLogin()
 			user.Type = model.NormalUserType
 			user.AvatarURL = ghUser.GetAvatarURL()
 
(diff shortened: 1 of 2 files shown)


vinamra28:
@kk-r can you please squash the commits as well?

kk-r:
@vinamra28 Sure, will do!

api/pkg/service/auth/auth.go
```diff
@@ -123,7 +123,7 @@ func (r *request) addUser(ghUser *github.User) (*model.User, error) {
 
 	// Check if user exist
 	q := r.db.Model(&model.User{}).
-		Where("LOWER(github_login) = ?", strings.ToLower(ghUser.GetLogin()))
+		Where("LOWER(github_login) = ? or github_login = ?", strings.ToLower(ghUser.GetLogin()), ghUser.GetLogin())
```
sm43, line 126: here in the first condition we are converting to lowercase and then checking
`LOWER(github_login) = ?", strings.ToLower(ghUser.GetLogin())`
so would we require the 2nd check after `or` ?
first one will cover all cases I think as it is converting to lower case both the words
wdyt?

kk-r: @sm43 Thanks. Yeah. That's make sense. Sorry, it was added as part of initial plan. It's not required anymore. TBC, this logic `LOWER(github_login) = ?", strings.ToLower(ghUser.GetLogin())` will be sufficient to handle all the scenarios mentioned below.

```
In DB                       From GH
username small case   == username small case
username mixed case   == username mixed case
username small case   == username mixed case
username mixed case   == username small case

sm43:
Thanks :)
/approve
/meow

vinamra28 (approved):
/lgtm

PR #306, opened 2021-08-10
 Updates auth UI to supports multiple auth provider

pratap0007:
# Changes
- Updates Auth UI to supports multiple auth providers

Screenshot from 2021-12-09 10-35-06

Screenshot from 2021-12-08 16-21-18

- While doing login if status code isn't 200 then it shows warning message on ui 

Screenshot from 2022-01-05 10-29-48

# Submitter Checklist

review them:


ui/src/containers/Header/index.tsx
```diff
@@ -82,6 +82,33 @@ const Header: React.FC = observer(() => {
...
+          </Button>
+        ]}
+      >
+        <Grid>
+          <GridItem style={{ marginBottom: '2em' }}>
+            <Button> Sign In with GitHub Enterprise</Button>
+          </GridItem>
+          <GridItem style={{ marginBottom: '2em' }}>
+            <Button> Sign In with GitHub</Button>
+          </GridItem>
```
vinamra28, line 103: I guess we should keep only `Sign In with Github`

ui/src/containers/Header/index.tsx
```diff
@@ -82,6 +82,33 @@ const Header: React.FC = observer(() => {
...
+          <GridItem style={{ marginBottom: '2em' }}>
+            <Button> Sign In with GitLab</Button>
+          </GridItem>
+          <GridItem style={{ marginBottom: '2em' }}>
+            <Button> Sign in with GitBucket</Button>
```
vinamra28, line 108: nit:
```suggestion
            <Button> Sign in with BitBucket</Button>
```

ui/src/containers/Header/index.tsx
pratap0007: You mean , we don't need to keep  `Sign In with GitHub Enterprise` separately

vinamra28: yeah... either should be there

sm43:
is that button at centre 😅 seems a bit not aligned
and keep the close button a little right corner if possible ?

ui/src/store/auth.ts
```diff
@@ -120,13 +120,13 @@ export const AuthStore = types
...
         }
       } catch (err) {
-        const error: IError = {
-          status: err.status,
-          serverMessage: titleCase(err.data.message),
-          customMessage: ''
-        };
-        self.setErrorMessage(error);
-        self.authErr = error;
+        // const error: IError = {
```
PuneetPunamiya: Do we need this ??

ui/src/containers/Header/index.tsx
```diff
@@ -82,6 +90,45 @@ const Header: React.FC = observer(() => {
...
+                Close
+              </Button>
+            </GridItem>
+          </Grid>
+        ]}
+      >
+        <Grid>
+          {providers.values.slice(0, 1).map((provider: IProvider) => (
+            <GridItem key={provider.name} offset={3} span={6} className="header-sigin-button">
+              <Button
```
PuneetPunamiya, line 107: I just had a question, if the button is clicked once is the button disabled ?
Also what are we doing if there is some error coming from the auth server, are we setting any alerts ??

pratap0007:
> is that button at centre sweat_smile seems a bit not aligned and keep the close button a little right corner if possible ?

button is at center  , yes, will  try to keep right corner

ui/src/containers/Header/index.tsx
pratap0007: it does't disable ,there is no issue if it is clicked more than once

ui/src/store/auth.ts
pratap0007: yes

ui/src/components/Icon/Icon.test.tsx
```diff
@@ -43,4 +43,9 @@ describe('Icon Component', () => {
     expect(component.debug()).toMatchSnapshot();
     expect(component.find('GithubIcon[label="Github"]').length).toEqual(1);
   });
+  it('should render icon for Gitlab', () => {
+    const component = shallow(<Icon id={Icons.Gitlab} size="sm" label="Gitlab" />);
+    expect(component.debug()).toMatchSnapshot();
+    expect(component.find('GitlabIcon[label="Gitlab"]').length).toEqual(1);
+  });
```
PuneetPunamiya, line 50: I think we are missing test for bitbucket icon

ui/src/containers/Header/Header.css
```diff
@@ -8,3 +8,11 @@
   margin-left: 0.2em;
   cursor: pointer;
 }
+
+.header-sigin-button {
```
PuneetPunamiya, line 12: I think we should have the prefix as `hub-`

ui/src/containers/Header/Header.css
```diff
@@ -8,3 +8,11 @@
...
+  margin-bottom: 2em;
+}
+
+.header-sigin-button__icon {
```
PuneetPunamiya, line 16: same as above

ui/src/store/auth.ts
PuneetPunamiya: You mean comments ?

ui/src/containers/Header/Header.css
```diff
@@ -10,9 +10,50 @@
...
+.header-sigin-button a:hover svg {
+  color: white;
+}
+
+.header-login__modal h6 {
+  margin-bottom: -0.5em !important;
+  text-align: center;
+}
+
+.pf-c-backdrop {
```
PuneetPunamiya, line 57: I think we should not modify the root component

PuneetPunamiya: May be a component which we are adding inside that we can add a className and then modify it

ui/src/containers/Header/Header.css
```diff
@@ -10,9 +10,50 @@
...
+  color: white;
+}
+
+.header-sigin-button a:hover svg {
+  color: white;
+}
+
+.header-login__modal h6 {
+  margin-bottom: -0.5em !important;
+  text-align: center;
```
PuneetPunamiya, line 54: Could you please change the name of the classes with prefix as `hub-`

ui/src/store/auth.ts
```diff
@@ -120,13 +125,17 @@ export const AuthStore = types
...
-        // self.authErr = error;
+        if (err === undefined) {
+          self.isAuthenticated = false;
+        } else {
+          const error: IError = {
+            status: err.status,
+            serverMessage: titleCase(err.data),
+            customMessage: ''
+          };
+          self.setErrorMessage(error);
```
PuneetPunamiya, line 136: Should we create a function for this and reuse it is other actions ?

ui/src/containers/ParseUrl/index.tsx
```diff
@@ -1,30 +1,40 @@
...
+  const { resources, user } = useMst();
+  const history = useHistory();
 
   if (window.location.search) {
     const searchParams = new URLSearchParams(window.location.search);
+    const status = searchParams.get('status');
+    const code = searchParams.get('code');
+
+    // It checks status and code and then redirect to authentication
+    if (status === '200' && code !== null) {
```
PuneetPunamiya, line 17: What are we doing if status is not 200 ?

PuneetPunamiya: Showing the error is alert message ?

ui/src/containers/Header/Header.css
pratap0007: yes, but in hub we are using patternfly modal in two components and if we want to blur the background (as require same css fields) then we need at root level css otherwise we need to define same css in both component css file where patternfly modal have used

ui/src/containers/ParseUrl/index.tsx
pratap0007: In case status is not 200 then user needs to retry login again and we are not showing any error or alert message

PuneetPunamiya: I think we should some error in the form of alert in that case so that it becomes easier for the user to understand

ui/src/store/auth.ts
```diff
@@ -120,13 +120,18 @@ export const AuthStore = types
...
-        self.authErr = error;
+        if (err === undefined) {
+          self.isAuthenticated = false;
+        } else {
+          const error = createError(err.status, err.data, '');
+          // const error: IError = {
+          //   status: err.status,
+          //   serverMessage: titleCase(err.data),
+          //   customMessage: ''
+          // };
```
vinamra28, line 131: can we remove this?

vinamra28 (commented):
/lgtm

ui/src/containers/Header/Header.css
PuneetPunamiya: May be then where we are using the component there we can give inline styles

ui/src/containers/Header/Header.test.tsx
```diff
@@ -72,19 +84,30 @@ describe('Header', () => {
...
     };
     user.authenticate(code);
-    user.updateRefreshToken();
     when(
       () => {
         return !user.isLoading;
       },
       () => {
         setTimeout(() => {
+          component.debug();
```
PuneetPunamiya, line 93: Do we need this ?

ui/image/start.sh
```diff
@@ -13,9 +13,7 @@ echo '----------------------------------------------'
 cat <<-EOF > $CONFIG_JS
 window.config = {
   API_URL: '$API_URL',
-  GH_CLIENT_ID: '$GH_CLIENT_ID',
```
vinamra28, line 16: add `REDIRECT_URI` here as well

ui/src/config/constants.tsx
```diff
@@ -1,9 +1,9 @@
 window.config = window.config || {
   API_URL: 'no API_URL  set',
-  GH_CLIENT_ID: 'no GH_CLIENT_ID set'
+  AUTH_BASE_URL: 'no AUTH_BASE_URL set'
```
vinamra28, line 3: don't we need to add `REDIRECT_URI` here as well?

ui/src/containers/Header/Header.test.tsx
pratap0007: nope, we can remove it

ui/src/config/constants.tsx
pratap0007: not needed, but it's good to add and I will add for that

vinamra28 (commented):
/lgtm

PR #307, opened 2021-08-11
Add platforms support to hub cli

barthy1:
# Changes

- add platforms option to search
- show platforms in the resource info (for the resource version)
- show platforms in search results (for the resource)

# Submitter Checklist

review them:


api/pkg/service/resource/resource.go
```diff
@@ -60,6 +60,7 @@ func (s *service) Query(ctx context.Context, p *resource.QueryPayload) (*resourc
 		Catalogs:   p.Catalogs,
 		Categories: p.Categories,
 		Tags:       p.Tags,
+		Platforms:  p.Platforms,
```
sm43, line 63: could you also add this field for /v1 api?

barthy1: @sm43 the Platforms field is added to /v1 api.

api/pkg/cli/formatter/field.go
```diff
@@ -73,6 +74,22 @@ func FormatTags(tags []*client.TagResponseBody) string {
 	return sb.String()
 }
 
+// FormatPlatforms returns list of platforms seperated by comma
+func FormatPlatforms(platforms []*client.PlatformResponseBody) string {
+	var sb strings.Builder
+	if len(platforms) == 0 {
```
PuneetPunamiya, line 80: Just a thought, as according to the [proposal](https://github.com/tektoncd/community/blob/main/teps/0070-tekton-catalog-task-platform-support.md#proposal) if the annotation for platform is not specified, `linux/amd64` platform is meant by default right, should we add `linux/amd64` instead of `---`

barthy1: hmm, interesting idea, actually. At this moment this logic (no platform = `linux/amd64` platform) is implemented on hub resource source code parsing level, so expectation is that all resources in DB already have `linux/amd64` if their source code doesn't have platform annotation. 
So I'd keep it as is, because substitution logic is already implemented on the backend. And if there are no platforms, it means that hub deployment doesn't have latest code and showing `linux/amd64` for all tasks can be incorrect information.  What do you think?

PuneetPunamiya: Yes I think let's keep it from what we get from backend instead on manually adding it, since resources with no platform already have `linux/amd64` in db, so the above implementation looks fine to me.
Thanks

sm43:
As part of `search` cmd output we need to think what do we want to show 
```
NAME            KIND       CATALOG   DESCRIPTION                                  TAGS         PLATFORMS
```
#311 add `categories` column in this search result
each record might go to 2nd line when displayed on terminal, as now there are more number of columns 
if it goes to next line it would be difficult to understand the data.

@piyush-garg @vinamra28 @pratap0007 @PuneetPunamiya

PuneetPunamiya:
@barthy1  Could you please rebase the PR

pratap0007:
@barthy1, Just have a thought for install,upgrade and downgrade cli command, if user install/upgrade/downgrade a task that differs in supported platform from user system supported architecture and in that case it would fail, so my though is show an warning and abort the install/upgrade/downgrade of that task ,WDYT ?
cc: @piyush-garg @sm43 @PuneetPunamiya @vinamra28

piyush-garg:
@barthy1 

So we will be adding the platforms, categories and tags behind the `-o wide` flag. PR #311 add the `-o wide` flag, after that we can have this PR in `-o wide` flag.

barthy1:
@pratap0007 
> Just have a thought for install,upgrade and downgrade cli command, if user install/upgrade/downgrade a task that differs in supported platform from user system supported architecture and in that case it would fail, so my though is show an warning and abort the install/upgrade/downgrade of that task ,WDYT ?

it's really tricky :( the cli platform can differ from cluster platform. For instance, you have cli on your laptop (linux/amd64 or even windows/amd64), but your remote cluster runs on linux/ppc64le or linux/s390x or even heterogeneous (several platforms). So cli platform check is not the valid verifier here.

pratap0007:
> it's really tricky :( the cli platform can differ from cluster platform. For instance, you have cli on your laptop (linux/amd64 or even windows/amd64), but your remote cluster runs on linux/ppc64le or linux/s390x or even heterogeneous (several platforms). So cli platform check is not the valid verifier here.

@barthy1 Yeah,agree with you and thanks for clarifying it

api/pkg/cli/cmd/search/search.go
```diff
@@ -156,7 +160,7 @@ func (opts *options) run() error {
 func (opts *options) validate() error {
 
 	if flag.AllEmpty(opts.args, opts.kinds, opts.tags, opts.categories) {
```
pratap0007, line 162: Could you please add platforms here as well to check emptiness of platforms flag
```suggestion
	if flag.AllEmpty(opts.args, opts.kinds, opts.tags, opts.categories,opts.platforms) {
```

api/pkg/cli/cmd/search/search.go
```diff
@@ -31,19 +31,19 @@ import (
...
+{{ formatName $r.Name $r.LatestVersion.Version }}	{{ $r.Kind }}	{{ formatCatalogName $r.Catalog.Name }}	{{ formatDesc $r.LatestVersion.Description 40 }}	{{ formatPlatforms $r.LatestVersion.Platforms }}	{{ formatTags $r.Tags }}	{{ formatCategories $r.Categories }}
 {{ end }}
 {{- end -}}
 `
 
 const minResTemplate = `{{- $rl := len .Resources }}{{ if eq $rl 0 -}}
 No Resources found
 {{ else -}}
-NAME	KIND	CATALOG	DESCRIPTION
+NAME	KIND	CATALOG	DESCRIPTION	PLATFORMS
```
pratap0007, line 44: Could you please remove the `PLATFORMS` as recently we had a discussion to keep  the Tags,Categories and Platforms  with `-o wide` flag only

pratap0007:
Thanks for updating the PR and it looks good to me
cc: @PuneetPunamiya @piyush-garg

PR #359, opened 2022-01-31
Handles user data username case formatting

PuneetPunamiya:
- While running the api, config.yaml has some entries of username
    with some scopes added, which are stored in db, so there can be a
    scenario where user's username case is different than what it is
    actually on the Git application. This might loss the user's existing
    data i.e. the rating of resources which user has already given
    because it would create a new entry if username won't be found.

  - Hence this patch checks for the case of username while inserting
    the user data when user authenticates to Hub.

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/auth/service/service.go
+++ b/api/pkg/auth/service/service.go
@@ -17,6 +17,7 @@ package auth
 import (
 	"crypto/sha256"
 	"encoding/hex"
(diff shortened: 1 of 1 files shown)


vinamra28 (commented):
/lgtm

PR #435, opened 2022-04-07
Redirect preview API's to `/v1`

PuneetPunamiya:
* This patch redirects following preview API's to `/v1`

  * **VersionsByID** - `/resource/{id}/versions` --> `/v1/resource/{id}/versions`
  
  * **ByCatalogKindNameVersion** - `/resource/{catalog}/{kind}/{name}/{version}` --> `/v1/resource/{catalog}/{kind}/{name}/{version}`
  
 
  * **ByVersionID** - /resource/version/{versionID}` -->  `/v1/resource/version/{versionID}`
  
  * **ByCatalogKindName** - `/resource/{catalog}/{kind}/{name}` --> `/v1/resource/{catalog}/{kind}/{name}`
  
  * **ByResourceID** - `/resource/{id}` --> `/v1/resource/{id}`
  
  * **Query** - `/query` --> `/v1/query`

# Changes

# Submitter Checklist

review them:


vdemeester (approved):

piyush-garg:
lgtm
/cc @sm43 @vinamra28

api/pkg/service/resource/resource.go
```diff
@@ -17,98 +17,133 @@ package resource
...
-			Name: platform.Name,
-		})
+	for _, v := range res {
+		query = fmt.Sprintf("%s%v", query, v)
 	}
-	res.Platforms = platforms
 
-	res.HubURLPath = fmt.Sprintf("%s/%s/%s/%s", r.Resource.Catalog.Name, r.Resource.Kind, r.Resource.Name, r.Version)
+	query = strings.Replace(query, "&", "", 1)
+	fmt.Println(query)
```
vinamra28, line 103: nit: can you please remove this?

vinamra28 (approved):
just one minor nitpick else looks good

api/pkg/service/resource/resource.go
```diff
@@ -17,98 +17,132 @@ package resource
...
 	}
-	rv.Latest = minVersionInfo(versions[len(versions)-1])
 
-	return &resource.ResourceVersions{Data: &rv}, nil
-}
+	if p.Tags != nil {
+		res["Tags"] = fmt.Sprintf("&tags=%s", p.Tags[0])
+		if len(p.Tags) > 1 {
+			for i := 1; i < len(p.Tags); i++ {
+				res["Tags"] = fmt.Sprintf("%v%s", res["Tags"], fmt.Sprintf("&tags=%s", p.Tags[i]))
```
pratap0007, line 76:
```suggestion
		res["Tags"] = fmt.Sprintf(strings.Join(p.Tags,"&tags="))
```

api/pkg/service/resource/resource.go
```diff
@@ -17,98 +17,132 @@ package resource
...
+			}
+		}
+	}
 
-func tinyVersionInfo(r model.ResourceVersion) *resource.ResourceVersionData {
+	if p.Platforms != nil {
+		res["Platforms"] = fmt.Sprintf("&platforms=%s", p.Platforms[0])
```
pratap0007, line 82: and we can do same here

api/pkg/service/resource/resource.go
```diff
@@ -17,98 +17,132 @@ package resource
...
-	res.WebURL = r.URL
-	res.RawURL = getStringReplacer(r.URL, r.Resource.Catalog.Provider).Replace(r.URL)
-	platforms := []*resource.Platform{}
-	for _, platform := range r.Platforms {
-		platforms = append(platforms, &resource.Platform{
-			ID:   platform.ID,
-			Name: platform.Name,
-		})
+	for _, v := range res {
+		query = fmt.Sprintf("%s%v", query, v)
```
pratap0007, line 99: we are using map for query params construction , query params order gets change sometime after redirection e.g, `/query?tags=aws&tags=acr&limit=1` --> `/v1/query?limit=1&tags=aws&tags=ecr`.
I don't think it could be any issue , wdyt?

pratap0007:
lgtm

api/pkg/service/resource/resource.go
PuneetPunamiya: yeah I don't think it would be any issue since the response would be the same

api/pkg/service/resource/resource.go
```diff
@@ -17,98 +17,107 @@ package resource
...
 func New(api app.BaseConfig) resource.Service {
 	return &service{api.Service("resource")}
 }
 
-// List all resources sorted by rating and name
-func (s *service) List(ctx context.Context) (*resource.Resources, error) {
-	return &resource.Resources{}, nil
-}
+// Find resources based on name, kind or both
+func (s *service) Query(ctx context.Context, p *resource.QueryPayload) (*resource.QueryResult, error) {
```
piyush-garg, line 36: This func can be refactored/improved a bit but can handle in a follow-up PR

api/design/resource.go
```diff
@@ -38,21 +96,212 @@ var _ = Service("resource", func() {
...
+				Header("location")
+			})
 
-			Response(StatusOK)
-			Response("internal-error", StatusInternalServerError)
-			Response("not-found", StatusNotFound)
+		})
+	})
+
+	Method("ByCatalogKindNameVersionYaml", func() {
```
sm43, line 191: not a preview api 
we can skip adding this

piyush-garg:
We dont need to set redirect for /readme and /yaml apis as they were not part of preview apis before

api/pkg/service/resource/resource.go
```diff
@@ -17,98 +17,120 @@ package resource
...
-	res.HubURLPath = fmt.Sprintf("%s/%s/%s/%s", r.Resource.Catalog.Name, r.Resource.Kind, r.Resource.Name, r.Version)
+// Find a resource using it's id
+func (s *service) ByID(ctx context.Context, p *resource.ByIDPayload) (*resource.ByIDResult, error) {
+	return &resource.ByIDResult{Location: fmt.Sprintf("/v1/resource/%d", p.ID)}, nil
+}
 
-	return res
+func queryBuilder(queryParam, key, value string) string {
+	if len(queryParam) == 0 {
+		return queryParam + fmt.Sprintf("%s=%s", key, value)
```
sm43, line 133: we can use
```
v := url.Values{}
v.Set()
```
a bit cleaner

sm43: (could be a follow up pr)

api/design/resource.go
```diff
@@ -38,21 +96,212 @@ var _ = Service("resource", func() {
...
+			})
+			Required("location")
+		})
+
+		HTTP(func() {
+			GET("/resource/{catalog}/{kind}/{name}")
+
+			Param("pipelinesversion")
+
+			Response(StatusFound, func() {
```
sm43, line 278: just trying to understand
so we add the url we want to redirect in response header as `Location`
and goa will redirect to that API?

PuneetPunamiya: yes so we are adding an attribute in `Result` as being the value for the location header and then we map that to the Location response header to which it redirects to the value which we set in the service logic

api/pkg/service/resource/resource.go
PuneetPunamiya: yeah sure will address it in a follow up pr

PR #510, opened 2022-05-25
Add resources to database during api start

pratap0007:
This patch adds  following 
- Adds an `apiserver` user with catalog refresh scope
and adds resources from catalogs to the database only during
first time api start

- Now user don't need to login on Hub ui inorder to add
 resources to db from catalog, it would add automatically
 during api start so this commit updates docs accordingly

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/api/cmd/api/main.go
+++ b/api/cmd/api/main.go
@@ -25,14 +25,6 @@ import (
 	"strings"
 	"sync"
 
-	// Go runtime is unaware of CPU quota which means it will set GOMAXPROCS
-	// to underlying host vm node. This high value means that GO runtime
-	// scheduler assumes that it has more threads and does context switching
-	// when it might work with fewer threads.
-	// This doesn't happen# with our other controllers and services because
-	// sharedmain already import this package for them.
-	_ "go.uber.org/automaxprocs"
-
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	admin "github.com/tektoncd/hub/api/gen/admin"
@@ -56,6 +48,14 @@ import (
 	v1resource "github.com/tektoncd/hub/api/v1/gen/resource"
 	v1catalogsvc "github.com/tektoncd/hub/api/v1/service/catalog"
 	v1resourcesvc "github.com/tektoncd/hub/api/v1/service/resource"
+
+	// Go runtime is unaware of CPU quota which means it will set GOMAXPROCS
+	// to underlying host vm node. This high value means that GO runtime
+	// scheduler assumes that it has more threads and does context switching
+	// when it might work with fewer threads.
+	// This doesn't happen# with our other controllers and services because
+	// sharedmain already import this package for them.
+	_ "go.uber.org/automaxprocs"
 )
 
 func main() {
@@ -90,6 +90,12 @@ func main() {
 	initializer := initializer.New(api)
 	if _, err := initializer.Run(context.Background()); err != nil {
 		logger.Fatalf("Failed to populate table: %v", err)
+	}
+
+	// Add resources to database
+	db := initializer.DB(context.Background())
+	if err := initializer.AddResources(db, api, logger); err != nil {
+		logger.Fatalf("Failed to add resources: %v", err)
 	}
 
 	// Initialize the services.
--- a/api/pkg/db/initializer/initializer.go
+++ b/api/pkg/db/initializer/initializer.go
@@ -21,6 +21,7 @@ import (
 	"github.com/tektoncd/hub/api/gen/log"
 	"github.com/tektoncd/hub/api/pkg/app"
 	"github.com/tektoncd/hub/api/pkg/db/model"
+	catalogsvc "github.com/tektoncd/hub/api/pkg/service/catalog"
 	"gorm.io/gorm"
 )
 
@@ -30,6 +31,13 @@ type Initializer struct {
 	app.Service
 	api app.BaseConfig
 }
+
+var (
+	apiserverID       = 0xFFFF_FFFF
+	apiserverName     = "api-server-bot"
+	apiserverUserName = "apiserver-bot"
+	apiserverEmailID  = "apiserver-bot"
+)
 
 // New returns the Initializer implementation.
 func New(api app.BaseConfig) *Initializer {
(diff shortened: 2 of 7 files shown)


api/pkg/db/initializer/initializer.go
```diff
@@ -31,6 +32,13 @@ type Initializer struct {
 	api app.BaseConfig
 }
 
+var (
+	apiserverID       = 100000
+	apiserverName     = "apiserver"
+	apiserverUserName = "apiserver"
+	apiserverEmailID  = "apiserver@gmail.com"
```
sm43, line 39:
```suggestion
	apiserverEmailID  = "apiserver"
```

api/pkg/db/initializer/initializer.go
```diff
@@ -31,6 +32,13 @@ type Initializer struct {
 	api app.BaseConfig
 }
 
+var (
+	apiserverID       = 100000
+	apiserverName     = "apiserver"
```
sm43, line 37:
```suggestion
	apiserverName     = "api-server"
```

api/pkg/db/initializer/initializer.go
```diff
@@ -196,3 +204,84 @@ func withTransaction(db *gorm.DB, log *log.Logger, data *app.Data, fns ...initFn
...
+	}
+
+	// Adds catalog refresh scope to user apiserver
+	scope := model.Scope{}
+	db.Where("name = ?", "catalog:refresh").Find(&scope)
+
+	apiUserScope := model.UserScope{
+		UserID:  uint(apiserverID),
+		ScopeID: scope.ID,
+	}
```
vinamra28, line 228: do we need to give bot the scopes ?

api/pkg/db/initializer/initializer.go
```diff
@@ -31,6 +32,13 @@ type Initializer struct {
 	api app.BaseConfig
 }
 
+var (
+	apiserverID       = 100000
+	apiserverName     = "api-server"
+	apiserverUserName = "apiserver"
+	apiserverEmailID  = "apiserver"
```
vinamra28, line 39: probably we can append `-bot` in the suffix so that searching for a bot account is easy. WDYT?

api/pkg/db/initializer/initializer.go
```diff
@@ -196,3 +204,84 @@ func withTransaction(db *gorm.DB, log *log.Logger, data *app.Data, fns ...initFn
...
+
+	if err := db.Create(&apiAccount).Error; err != nil {
+		log.Error(err)
+		return err
+	}
+
+	return nil
+}
+
+//AddResources func creates an apiserver account and adds resources from catalog to database
```
vinamra28, line 251: nit
```suggestion
// AddResources func creates an apiserver account and adds resources from catalog to database
```

api/cmd/api/main.go
```diff
@@ -92,7 +91,12 @@ func main() {
 		logger.Fatalf("Failed to populate table: %v", err)
 	}
 
-	// Initialize the services.
```
vinamra28, line 95: I guess we can still keep this comment

api/pkg/db/initializer/initializer.go
pratap0007: No, needs to remove

api/pkg/db/initializer/initializer.go
```diff
@@ -196,3 +204,70 @@ func withTransaction(db *gorm.DB, log *log.Logger, data *app.Data, fns ...initFn
 	txn.Commit()
 	return nil
 }
+
+func addApiServerUser(db *gorm.DB, log *log.Logger) error {
+
+	newUser := model.User{
+		Type:  "user",
```
vinamra28, line 211: I wonder if this should of type agent instead of user 🤔

api/cmd/api/main.go
```diff
@@ -31,8 +31,6 @@ import (
 	// when it might work with fewer threads.
 	// This doesn't happen# with our other controllers and services because
 	// sharedmain already import this package for them.
-	_ "go.uber.org/automaxprocs"
```
piyush-garg, line 34: Move the comments also please

api/pkg/db/initializer/initializer.go
```diff
@@ -31,6 +32,13 @@ type Initializer struct {
 	api app.BaseConfig
 }
 
+var (
+	apiserverID       = 100000
```
piyush-garg, line 36: can we give some id which cannot be there in current already deployed hub db - something like zero etc

api/pkg/db/initializer/initializer.go
```diff
@@ -196,3 +204,70 @@ func withTransaction(db *gorm.DB, log *log.Logger, data *app.Data, fns ...initFn
...
+
+	if err := q.First(&apiAccount).Error; err == gorm.ErrRecordNotFound {
+
+		logger.Infof("user %s account not found,creating account for apiserver", "apiserver")
+
+		if err := addApiServerUser(db, logger); err != nil {
+			logger.Error(err)
+			return err
+		}
+	}
```
PuneetPunamiya, line 253: I think `AddResources` function does lot more than it is supposed to do, so we can move this logic into a separate function as it is checking if the user exists, if not it adds an `apiServer` user

In short we can add a function something like this 

```golang
func checkIfApiServerAccountExists(db *gorm.DB, logger *log.Logger) error {
	// Check if apiserver account exists or not
	// If not, it creates apiserver account with catalog refresh scope
	q := db.Model(&model.Account{}).Where("user_name = ?", apiserverUserName)
	apiAccount := model.Account{}

	if err := q.First(&apiAccount).Error; err == gorm.ErrRecordNotFound {

		logger.Infof("user %s account not found,creating account for apiserver", "apiserver")

		if err := addApiServerUser(db, logger); err != nil {
			logger.Error(err)
			return err
		}
	}

	return nil
}

func (i *Initializer) AddResources(db *gorm.DB, api app.BaseConfig, logger *log.Logger) error {

	if err := checkIfApiServerAccountExists(db, logger); err != nil {
		return err
	}
        
        ... 
        other part of the logic....
        ......
}

```

api/pkg/db/initializer/initializer.go
```diff
@@ -196,3 +204,70 @@ func withTransaction(db *gorm.DB, log *log.Logger, data *app.Data, fns ...initFn
...
+
+	return nil
+}
+
+// AddResources func creates an apiserver account and adds resources from catalog to database
+func (i *Initializer) AddResources(db *gorm.DB, api app.BaseConfig, logger *log.Logger) error {
+
+	// Check if apiserver account exists or not
+	// If not it creates apiserver account with catalog refresh scope
```
PuneetPunamiya, line 241: nit
```suggestion
	// If user does not exists, it creates apiserver account with catalog refresh scope
```

api/pkg/db/initializer/initializer.go
pratap0007: it' s not working with id value zero or negative, can you please suggest any other id ?

vinamra28: since we are using uint so negative integers are not allowed, how about we use it's max possible value `4,29,49,67,295` ?

piyush-garg:
@pratap0007 Don't we need to update the docs also accordingly. Will it work for private catalogs too?

pratap0007:
> @pratap0007 Don't we need to update the docs also accordingly. Will it work for private catalogs too?

yes it will work for private catalog and for docs we will need to update it , shall I update the docs in this PR or in a new PR?

docs/DEPLOYMENT.md
```diff
@@ -381,50 +378,10 @@ For `OpenShift`:-
 
 Open the URL in a browser.
 
-Note: In UI you won't be able see the resources on homepage as there are not any in the db but you should see Categories on the left side.
+Note: In UI you should able to see the resources on homepage and Categories on the left side.
 
 Update your GitHub OAuth created with the UI route in place of `Homepage URL` and `Authorization callback URL`.
 
-## Add resources in DB
```
PuneetPunamiya, line 388: I think we should not delete this, rather change the heading because if someone wants to update the catalog he needs to refresh the catalog

docs/DEPLOYMENT.md
```diff
@@ -292,9 +291,7 @@ For `OpenShift`:-
 curl -k -X GET $(oc get -n tekton-hub routes api --template='https://{{ .spec.host }}/v1/categories')
 ```
 
-NOTE: At this moment, there are no resources in the db. Only the categories from hub config are added in the db.
-
-Let's deploy the UI first and then we will add the resources in db.
+NOTE: At this moment, resources are added in the db from the catalogs also categories from hub config are added in the db.
```
vinamra28, line 294: do we need this `NOTE` ?

docs/DEPLOYMENT.md
```diff
@@ -381,50 +378,10 @@ For `OpenShift`:-
 
 Open the URL in a browser.
 
-Note: In UI you won't be able see the resources on homepage as there are not any in the db but you should see Categories on the left side.
+Note: In UI you should able to see the resources on homepage and Categories on the left side.
```
vinamra28, line 381:
```suggestion
Note: On UI, you should be able to see the resources on homepage and Categories on the left side.
```

docs/DEPLOYMENT.md
pratap0007: I think we can remove it now cc: @PuneetPunamiya @piyush-garg

piyush-garg:
lgtm

PuneetPunamiya:
LGTM

vinamra28 (commented):
/lgtm
/meow

PR #693, opened 2022-10-20
Updates config file parsing and deployment files

pratap0007:
- Moves hub config data into the api ConfigMap
  and updates category format in the config file and in the configMap
- Updates api deployment to mount the configMap data
 - In case of development it parses config data from the local config file
 - In case of production it parses config data from the configMap mounted path

# Submitter Checklist

review them:

### DIFF
--- a/api/pkg/app/app.go
+++ b/api/pkg/app/app.go
@@ -83,9 +83,12 @@ type EnvMode string
 
 // Types of EnvMode
 const (
-	Production  EnvMode = "production"
-	Development EnvMode = "development"
-	Test        EnvMode = "test"
+	Production             EnvMode = "production"
+	Development            EnvMode = "development"
+	Test                   EnvMode = "test"
+	ConfigFileDir          string  = "/tmp/config/"
+	CatalogRefreshInterval string  = "catalog_refresh_interval"
+	DefaultScopes          string  = "default"
 )
 
 // DBDialect defines dialect for db connection
@@ -168,17 +171,47 @@ func (ab *APIBase) Data() *Data {
 
 // ReloadData reads config file and loads data in Data object
 func (ab *APIBase) ReloadData() error {
-	// Reads config file url from env
-	url, err := configFileURL()
-	if err != nil {
-		return err
-	}
 
-	// Reads data from config file
-	fileData, err := dataFromURL(url)
-	if err != nil {
-		ab.logger.Errorf("failed to read config file: %v", err)
-		return err
+	var fileData []byte
+
+	if ab.Environment() == Production {
+		// checks Config dir is present or not
+		dir, err := os.ReadDir(ConfigFileDir)
+		if err != nil {
+			ab.logger.Errorf("Directory %s not found : %s", ConfigFileDir, err)
+			return err
+		}
+		// Iterate on all keys to construct file of config data by adding key,value pair
+		for _, f := range dir {
+			//Skips to read hiden files
+			if !strings.Contains(f.Name(), "..") {
+				d, err := readLocalFile(ConfigFileDir + f.Name())
+				if err != nil {
+					return err
+				}
+				if strings.ToLower(f.Name()) == CatalogRefreshInterval {
+					continue
+				} else if strings.ToLower(f.Name()) == DefaultScopes {
+					fileData = append(fileData, []byte(fmt.Sprintf("\n%s:\n %s", strings.ToLower(f.Name()), string(d)))...)
+				} else {
+					fileData = append(fileData, []byte(fmt.Sprintf("\n%s:\n%s", strings.ToLower(f.Name()), string(d)))...)
+				}
+			}
+		}
+	} else {
+
+		url, err := configFileURL()
+		if err != nil {
+			return err
+		}
+
+		// Reads data from config file
+		data, err := dataFromURL(url)
+		if err != nil {
+			ab.logger.Errorf("failed to read config file: %v", err)
+			return err
+		}
+		fileData = append(fileData, data...)
 	}
 
 	// Viper unmarshals data from config file into Data Object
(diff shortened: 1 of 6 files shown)


config/02-api/21-api-configmap.yaml
```diff
@@ -19,5 +19,59 @@ metadata:
...
+      users: [vinamra28, piyush-garg, pratap0007, puneetpunamiya, sm43, sthaha, vdemeester]
+    - name: config:refresh
+      users: [vinamra28, piyush-garg, pratap0007, puneetpunamiya, sm43, sthaha, vdemeester]
+  DEFAULT: |
+    scopes:
+      - rating:read
+      - rating:write
   CATALOG_REFRESH_INTERVAL: 30m
+
+  # CATALOGS: |
```
PuneetPunamiya, line 63: Do we need this comments

api/pkg/app/app.go
```diff
@@ -83,9 +83,12 @@ type EnvMode string
 
 // Types of EnvMode
 const (
-	Production  EnvMode = "production"
-	Development EnvMode = "development"
-	Test        EnvMode = "test"
+	Production             EnvMode = "production"
+	Development            EnvMode = "development"
+	Test                   EnvMode = "test"
+	ConfigFileDir          string  = "/tmp/config/"
```
vinamra28, line 89: we'll probably have to mark this constant in the docs so that if somebody wants to deploy their hub then this value shouldn't change in deployment.yaml

api/pkg/app/app.go
```diff
@@ -188,13 +220,13 @@ func (ab *APIBase) ReloadData() error {
...
 		return err
 	}
+
 	if err := viper.Unmarshal(&data); err != nil {
 		ab.logger.Errorf("failed to unmarshal config data: %v", err)
 		return err
 	}
 	ab.data = data
 
-	// computes checksum on config data
```
vinamra28, line 197: any reason why this comment is removed?

api/pkg/app/app.go
```diff
@@ -168,17 +171,46 @@ func (ab *APIBase) Data() *Data {
...
+	if ab.Environment() == Production {
+		// checks Config dir is present or not
+		dir, err := os.ReadDir(ConfigFileDir)
+		if err != nil {
+			ab.logger.Errorf("Directory %s not found : %s", ConfigFileDir, err)
+			return err
+		}
+		// Iterate on all keys to construct file of config data by adding key,value pair
+		for _, f := range dir {
+			if !strings.Contains(f.Name(), "..") {
```
vinamra28, line 186: can we add a comment for this if statement?

api/pkg/app/app.go
pratap0007: no , will revert changes

api/pkg/app/app.go
```diff
@@ -168,17 +171,47 @@ func (ab *APIBase) Data() *Data {
...
+					return err
+				}
+				if strings.ToLower(f.Name()) == CatalogRefreshInterval {
+					continue
+				} else if strings.ToLower(f.Name()) == DefaultScopes {
+					fileData = append(fileData, []byte(fmt.Sprintf("\n%s:\n %s", strings.ToLower(f.Name()), string(d)))...)
+				} else {
+					fileData = append(fileData, []byte(fmt.Sprintf("\n%s:\n%s", strings.ToLower(f.Name()), string(d)))...)
+				}
+			}
```
PuneetPunamiya, line 199: In both the condition and else part we are adding the same logic, can we then skip the else if part then ?

config.yaml
```diff
@@ -13,26 +13,26 @@
...
-  - name: Kubernetes
-  - name: Messaging
-  - name: Monitoring
-  - name: Networking
-  - name: Openshift
-  - name: Publishing
-  - name: Security
-  - name: Storage
-  - name: Testing
+  - Automation
```
PuneetPunamiya, line 16: Shall we remove this file as we are adding all the data in the config map ?

PuneetPunamiya (commented):
Overall PR looks good to me, just some minor comments to address, else the patch is good to merge

api/pkg/app/app.go
pratap0007: in the else part it adds an extra white space

config.yaml
pratap0007: this is required to get configuration data in the local development , if we want to read configuration data from config map as well then I will remove it , wdyt? cc @piyush-garg @vinamra28

PuneetPunamiya: I think we should keep it then for the local development atleast for now

pratap0007:
could you please take a look on this PR @piyush-garg @sm43

api/pkg/app/app.go
```diff
@@ -83,9 +83,12 @@ type EnvMode string
...
 const (
-	Production  EnvMode = "production"
-	Development EnvMode = "development"
-	Test        EnvMode = "test"
+	Production             EnvMode = "production"
+	Development            EnvMode = "development"
+	Test                   EnvMode = "test"
+	ConfigFileDir          string  = "/tmp/config/"
+	CatalogRefreshInterval string  = "catalog_refresh_interval"
+	DefaultScopes          string  = "default"
```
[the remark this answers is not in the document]
PuneetPunamiya: I think yes we should make it to `Default`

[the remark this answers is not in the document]
pratap0007: we have a struct with same name `Default` here https://github.com/tektoncd/hub/blob/main/api/pkg/app/data.go#L52, so we can't define variable with same name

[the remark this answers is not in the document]
PuneetPunamiya: I'm fine with `DefaultScopes` @vinamra any thoughts on this ?

vinamra28 (commented):
/lgtm

PR #862, opened 2023-03-03
Remove lower case conversion of catalog name

pratap0007:
If catalog name is MyCatalg then clone path would be `tmp/catalog/MyCatalog` and file path generation was wrong due to lower conversion

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/api/v1/service/resource/resource.go
+++ b/api/v1/service/resource/resource.go
@@ -183,7 +183,7 @@ func (s *service) ByCatalogKindNameVersion(ctx context.Context, p *resource.ByCa
 func (s *service) ByCatalogKindNameVersionReadme(ctx context.Context,
 	p *resource.ByCatalogKindNameVersionReadmePayload) (*resource.ResourceVersionReadme, error) {
 
-	readmePath := fmt.Sprintf("%s/%s/%s/%s/%s/README.md", s.CatalogClonePath(), strings.ToLower(p.Catalog), strings.ToLower(p.Kind), p.Name, p.Version)
+	readmePath := fmt.Sprintf("%s/%s/%s/%s/%s/README.md", s.CatalogClonePath(), p.Catalog, strings.ToLower(p.Kind), p.Name, p.Version)
 	s.Logger(ctx).Info(fmt.Sprintf("Fetching README for resource %s", p.Name))
 	content, err := os.ReadFile(readmePath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


api/v1/service/resource/resource.go
```diff
@@ -183,7 +183,7 @@ func (s *service) ByCatalogKindNameVersion(ctx context.Context, p *resource.ByCa
 func (s *service) ByCatalogKindNameVersionReadme(ctx context.Context,
 	p *resource.ByCatalogKindNameVersionReadmePayload) (*resource.ResourceVersionReadme, error) {
 
-	readmePath := fmt.Sprintf("%s/%s/%s/%s/%s/README.md", s.CatalogClonePath(), strings.ToLower(p.Catalog), strings.ToLower(p.Kind), p.Name, p.Version)
+	readmePath := fmt.Sprintf("%s/%s/%s/%s/%s/README.md", s.CatalogClonePath(), p.Catalog, strings.ToLower(p.Kind), p.Name, p.Version)
```
vinamra28, line 186: instead of changing here, can't we convert the catalog name to lower case and then store or are we focusing on case sensitive?

api/v1/service/resource/resource.go
```diff
@@ -508,7 +508,7 @@ func versionInfoFromVersion(v model.ResourceVersion) *resource.ResourceVersion {
 func (s *service) GetRawYamlByCatalogKindNameVersion(ctx context.Context, p *resource.GetRawYamlByCatalogKindNameVersionPayload) (io.ReadCloser, error) {
 	s.Logger(ctx).Info(fmt.Sprintf("Fetching YAML for resource %s", p.Name))
 
-	yamlPath := fmt.Sprintf("%s/%s/%s/%s/%s", s.CatalogClonePath(), strings.ToLower(p.Catalog), strings.ToLower(p.Kind), p.Name, p.Version)
+	yamlPath := fmt.Sprintf("%s/%s/%s/%s/%s", s.CatalogClonePath(), p.Catalog, strings.ToLower(p.Kind), p.Name, p.Version)
```
vinamra28, line 511: also can we try to maintain consistency? I can see that in one of the changes done above, we have removed the lower case conversion for `p.Kind` and here we have kept that

api/v1/service/resource/resource.go
pratap0007: not sure on this, should we consider catalag name case sensitive? cc: @piyush-garg @PuneetPunamiya

jkandasa:
@pratap0007 can you please include bug id if you have any for this PR?

